PR #11, opened 2013-01-23
Support for boost filesystem v2

marioprats:


mirzashah:
Looks good, thanks!

mirzashah:
Just as an interesting note, apparently I removed v2 filesystem support when I was doing major refactoring to pluginlib. Did not realize folks were using older versions of boost.

PR #21, opened 2013-10-02
expose plugin paths in ClassLoader

dirk-thomas:
This enabled user land code to store the discovered plugin paths and later restore them to avoid timely recrawling.

@mirzashah @wjwwood Please review.

### DIFF
--- a/include/pluginlib/class_loader.h
+++ b/include/pluginlib/class_loader.h
@@ -61,9 +61,10 @@ namespace pluginlib
          * @param package The package containing the base class
          * @param base_class The type of the base class for classes to be loaded
          * @param attrib_name The attribute to search for in manifext.xml files, defaults to "plugin"
+         * @param plugin_xml_paths The list of paths of plugin.xml files, defaults to be crawled via ros::package::getPlugins()
          * @exception pluginlib::LibraryLoadException Thrown if package manifest cannot be found
          */
-        ClassLoader(std::string package, std::string base_class, std::string attrib_name = std::string("plugin"));
+        ClassLoader(std::string package, std::string base_class, std::string attrib_name = std::string("plugin"), std::vector<std::string> plugin_xml_paths = std::vector<std::string>());
 
         /**
          * @brief  Destructor for ClassLoader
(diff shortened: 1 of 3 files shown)


include/pluginlib/class_loader.h
```diff
@@ -104,6 +105,12 @@
          * @brief  Returns a list of all available classes for this ClassLoader's base class type
          * @return A vector of strings corresponding to the names of all available classes
          */
+        std::vector<std::string> getPluginXmlPaths();
```
wjwwood, line 108: Does the documentation need to be updated here?

include/pluginlib/class_loader.h
```diff
@@ -202,11 +209,19 @@
...
+        std::vector<std::string> getPluginXmlPaths(const std::string& package, const std::string& attrib_name);
+
+        /**
          * @brief  Returns the available classes
+         * @param plugin_xml_paths The vector of paths of plugin.xml files
          * @exception pluginlib::LibraryLoadException Thrown if package manifest cannot be found
          * @return A map of class names and the corresponding descriptions
          */
-        std::map<std::string, ClassDesc> determineAvailableClasses();
+        std::map<std::string, ClassDesc> determineAvailableClasses(const std::vector<std::string>& plugin_xml_paths);
```
wjwwood, line 224: Should there be (can there be) a version of this function which still doesn't take a parameter, or make this parameter have a default value? Otherwise we will be breaking API rather than just extending it.

include/pluginlib/class_loader_imp.h
```diff
@@ -50,7 +50,8 @@
 namespace pluginlib 
 {
   template <class T>
-  ClassLoader<T>::ClassLoader(std::string package, std::string base_class, std::string attrib_name) :
+  ClassLoader<T>::ClassLoader(std::string package, std::string base_class, std::string attrib_name, std::vector<std::string> plugin_xml_paths) :
```
wjwwood, line 53: Again, is it the intention to break API or just extend it? Maybe it doesn't matter.

include/pluginlib/class_loader.h
wjwwood, line 224: Nvm, this is private.

include/pluginlib/class_loader.h
dirk-thomas, line 108: Indeed.

include/pluginlib/class_loader_imp.h
dirk-thomas, line 53: This is only the _imp_ class. For the exposed ClassLoader constructor the new parameter is optional.

wjwwood, line 53: Yeah, I was just coming around to that understanding.

wjwwood:
+1

mirzashah:
Ah that's a nice feature. It's been a while since I've looked at this code...took me a while to get up to speed. Everything looks great, nice feature to have...will merge.

PR #24, opened 2014-04-21
update refreshDeclaredClasses to force recrawl (fix #23)

### LINKED ISSUE #23: refreshDeclaredClasses() should update chache
`refreshDeclaredClasses()` (http://docs.ros.org/hydro/api/pluginlib/html/class__loader__imp_8h_source.html#l00619) makes use of `ros::package` to find new plugins (see `getPluginXmlPaths`: http://docs.ros.org/hydro/api/pluginlib/html/class__loader__imp_8h_source.html#l00177). However, in the most cases newly available plugins are probably not yet in the cache. Hence `getPluginXmlPaths()` should use call `rospack profile` (or something similar) to update the cache before looking for new plugins.

dirk-thomas:
Depends on ros/ros#51.

### DIFF
--- a/include/pluginlib/class_loader.h
+++ b/include/pluginlib/class_loader.h
@@ -213,7 +213,7 @@ namespace pluginlib
          * @exception pluginlib::LibraryLoadException Thrown if package manifest cannot be found
          * @return A vector of paths
          */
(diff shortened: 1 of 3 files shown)


tfoote:
+1

esteve:
+1

bit-pirate:
Since this is not a breaking API change. Could it be released into hydro as well?

PR #26, opened 2014-05-08
Use cmake_modules to find TinyXML

### LINKED ISSUE #18: Use common FindTinyXML.cmake
There is a new package called cmake_modules, which we just created:

https://github.com/ros/cmake_modules

It provides a common FindTinyXML.cmake CMake module, which we should all try to use in order to prevent issues when doing things like cross-compiling.

This is a low priority, but a good idea. The package is available for Hydro and Groovy, it will be in public on the next sync.

esteve:
Fixes #18

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,22 +1,22 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(pluginlib)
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
+1 lgtm

PR #28, opened 2014-05-27
Make plugin_macro_update rosrunnable and remove it from global PATH

### LINKED ISSUE #12: plugin_macro_update probably should not be in global bin path
See also https://groups.google.com/d/topic/ros-sig-buildsystem/hsxeJocnKIQ/discussion

esteve:
Fixes #12


tfoote:
+1

PR #33, opened 2015-10-21
Add ":" to split function within getName.

mikeodr:
getName split wasn't supporting `::` as the delimiter for the package name and
the plugin name.

### DIFF
--- a/include/pluginlib/class_loader_imp.h
+++ b/include/pluginlib/class_loader_imp.h
@@ -385,7 +385,7 @@ namespace pluginlib
   {
     //remove the package name to get the raw plugin name
     std::vector<std::string> split;
-    boost::split(split, lookup_name, boost::is_any_of("/"));
+    boost::split(split, lookup_name, boost::is_any_of("/:"));
     return split.back();
   }
 


include/pluginlib/class_loader_imp.h
```diff
@@ -385,15 +385,15 @@ namespace pluginlib
   {
     //remove the package name to get the raw plugin name
     std::vector<std::string> split;
-    boost::split(split, lookup_name, boost::is_any_of("/"));
+    boost::split(split, lookup_name, boost::is_any_of("/:"));
```
mikeodr: This is the core of the change.

The rest is my editor stripping out excess white space.

tfoote:
Please remove the whitespace only changes. Although we agree with stripping them we don't want whitespace only changes applied in other patches, it makes the history unreadable (as well as the diff) and makes merges much harder. 

With respect to the content of the change what is your use case? It looks like you're changing the pre-groovy legacy support behavior for name lookup. [here](http://wiki.ros.org/pluginlib#pluginlib.2BAC8-pluginlib_groovy.Legacy_.22Lookup_Name.22)

paulbovbel:
We're looking to use getName() to extract a valid rosgraph name in both the legacy (package_name/plugin_name) and the modern (package_namespace::plugin_class) style.

mikeodr:
Whitespace removed.

esteve:
LGTM. Thanks @mikeodr !

PR #34, opened 2015-12-09
Fix wrong package name test

### LINKED ISSUE #31: failing unit test
http://jenkins.ros.org/job/devel-indigo-pluginlib/11/
http://jenkins.ros.org/job/devel-jade-pluginlib/3/

esteve:
Throw an exception if ClassLoader can't be instantiated due to an invalid package name

Fixes #31

### DIFF
--- a/include/pluginlib/class_loader_imp.h
+++ b/include/pluginlib/class_loader_imp.h
@@ -59,6 +59,11 @@ namespace pluginlib
   /***************************************************************************/
   {
     ROS_DEBUG_NAMED("pluginlib.ClassLoader","Creating ClassLoader, base = %s, address = %p", base_class.c_str(), this);
+    if (ros::package::getPath(package_).empty())
+    {
+      throw pluginlib::ClassLoaderException("Unable to find package: " + package_);
+    }
+
     if (plugin_xml_paths_.size() == 0)
     {
       plugin_xml_paths_ = getPluginXmlPaths(package_, attrib_name_);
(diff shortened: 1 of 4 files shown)


include/pluginlib/class_loader_imp.h
```diff
@@ -59,6 +59,11 @@ namespace pluginlib
   /***************************************************************************/
   {
     ROS_DEBUG_NAMED("pluginlib.ClassLoader","Creating ClassLoader, base = %s, address = %p", base_class.c_str(), this);
+    if (ros::package::getPath(package).empty())
```
dirk-thomas: Use `package_` instead of `package` as in the rest of the function body?

include/pluginlib/class_loader_imp.h
```diff
@@ -59,6 +59,11 @@ namespace pluginlib
...
+    {
+      throw pluginlib::ClassLoaderException("Unable to find package: " + package_);
```
dirk-thomas, line 64: The docblock needs to be updated for that.

include/pluginlib/class_loader_imp.h
esteve: Fixed.

include/pluginlib/class_loader_imp.h
esteve, line 64: Fixed.

dirk-thomas:
lgtm

dirk-thomas:
Does this rely on the change in rospack? If yes, this package needs to depend on the upcoming rospack version.

esteve:
It does not depend on https://github.com/ros/rospack/pull/51, the rospack PR is just a behavior we found out, but it's independent from this.

PR #35, opened 2015-12-26
Remove Boost Software License from license tag

jspricke:
The Boost Software License was only in there for Poco, as can be seen in
6e0659f. As Poco was removed in 44ab6fb and all remaining Files have a
BSD header, let's remove the Boost tag as well, to be consistent.


esteve:
Thanks!

PR #37, opened 2016-04-03
example illustrating two issues

rhaschke:
This isn't actually a PR, but a bug report. But the patch contains two example .cpp files illustrating the issues. IMHO the pluginlib doesn't work in cases where control over destruction of the ClassLoader is not possible. Typical use cases are e.g. liburdf to load urdf parsers.

issue1.cpp illustrates the use case I would expect:
In some function a `ClassLoader` is instantiated, some object from the lib is created and the ClassLoader is destroyed again. The destructor of ClassLoader tries to unload the lib, although the object is still in use. `libclass_loader` correctly notices the fact and denies unloading with a `SEVERE WARNING`.
IMHO, pluginlib's `ClassLoader` should simply use `libclass_loader's` `ondemand_load_unload` feature.

issue2.cpp illustrates actually an issue in `libclass_loader`:
As a work-around for issue1, people define their `ClassLoader` as `static`, thus postponing library unloading until program shutdown. However, as the order of releasing `static` variables is undefined, this fails, because an already freed static mutex is accessed:

```
issue2: /usr/include/boost/thread/pthread/recursive_mutex.hpp:110: void boost::recursive_mutex::lock(): Assertion `!pthread_mutex_lock(&m)' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff6450cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56  ../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff6450cc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff64540d8 in __GI_abort () at abort.c:89
#2  0x00007ffff6449b86 in __assert_fail_base (fmt=0x7ffff659a830 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x43a53e "!pthread_mutex_lock(&m)", 
    file=file@entry=0x43a508 "/usr/include/boost/thread/pthread/recursive_mutex.hpp", line=line@entry=110, 
    function=function@entry=0x43bba0 <boost::recursive_mutex::lock()::__PRETTY_FUNCTION__> "void boost::recursive_mutex::lock()") at assert.c:92
[...]

Issue #38, opened 2016-04-03
enabling on-demand loading/unloading fails


rhaschke:
To fix #37, I tried to enable on-demand loading/unloading. However, now classes cannot be instantiated:

```
terminate called after throwing an instance of 'pluginlib::CreateClassException'
  what():  MultiLibraryClassLoader: Could not create object of class type test_plugins::Foo as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
The program has unexpectedly finished.
```

rhaschke:
Will be fixed with #39.

PR #39, opened 2016-04-03
Enable on demand unloading

rhaschke:
Fixes #37, `issue1.cpp`. Relates to #38, which is fixed by https://github.com/ros/class_loader/pull/34.

### DIFF
--- a/include/pluginlib/class_loader_imp.h
+++ b/include/pluginlib/class_loader_imp.h
@@ -55,7 +55,7 @@ namespace pluginlib
   package_(package),
   base_class_(base_class),
   attrib_name_(attrib_name),
-  lowlevel_class_loader_(false) //NOTE: The parameter to the class loader enables/disables on-demand class loading/unloading. Leaving it off for now...libraries will be loaded immediately and won't be unloaded until class loader is destroyed or force unload.
+  lowlevel_class_loader_(true) // Enable on-demand class loading/unloading
   /***************************************************************************/
   {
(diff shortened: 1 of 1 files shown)


rhaschke:
The problem is definitely in `libclass_loader` only, which requires that the ClassLoaders stay in place as long as objects do.

rhaschke:
With https://github.com/ros/class_loader/pull/34 the unit tests should succeed and on-demand loading/unloading should work again.

mikaelarguedas:
Hi @rhaschke 
Given that https://github.com/ros/class_loader/pull/34 does fix the library loading issue for ondemand loading/unloading but doesn't solve the unloading problem when the `ClassLoader` is destroyed; 
Do you think that enabling ondemand loading/unloading in the pluginlib's ClassLoader still makes sense ? It will result in the same SEVERE WARNING and not unloading the library when ClassLoader destructor is called. As a result #37 `issue1.cpp` will not be fixed by this PR. Is that right?

rhaschke:
@mikaelarguedas 
Indeed, one gets the SEVERE WARNING when trying to destroy the ClassLoader when objects are still in use. Hence https://github.com/ros/class_loader/pull/34 doesn't fix both issues of #37.
I was thinking about to drop PR #39 too, because there is not benefit yet in using on-demand unloading of libraries: Either there is the SEVER WARNING or a segfault on exit of the program...
Hence, feel free to close this PR without merging.

mikaelarguedas:
@rhaschke I will then close this PR. The related issue can be kept open if anyone wants to take over and offer a solution to fix them.
Thanks for your work,

PR #41, opened 2016-09-11
Throw exception if plugin.xml is broken

cwecht:
I stumbled recently over an "basic_string::_M_construct null not valid" error using pluginlib. After about an hour of searching I found the problem: an attribute of an class tag was missing. I've created a test case to reproduce this error and now exceptions are thrown if required attributes of the class-tag are missing.

### DIFF
--- a/test/utest.cpp
+++ b/test/utest.cpp
@@ -184,12 +184,8 @@ TEST(PluginlibTest, brokenXML)
     SUCCEED();
     return;
   }
-  //catch(...)
-  //{
-  //  FAIL() << "Uncaught exception";
-  //}
-  ADD_FAILURE() << "Didn't throw exception as expected";
 
+  ADD_FAILURE() << "Didn't throw exception as expected";
 }
 
 // Run all the tests that were declared with TEST()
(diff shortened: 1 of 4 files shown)


test/utest.cpp
```diff
@@ -174,6 +173,25 @@ TEST(PluginlibTest, createManagedInstanceAndUnloadLibrary)
...
+  }
+  catch(pluginlib::ClassLoaderException& ex)
+  {
+    SUCCEED();
+    return;
+  }
+  //catch(...)
+  //{
+  //  FAIL() << "Uncaught exception";
+  //}
```
mikaelarguedas: these commented lines can be removed given that the test exits on line 185 if the expected exception is thrown

test/utest.cpp
```diff
@@ -174,6 +173,25 @@ TEST(PluginlibTest, createManagedInstanceAndUnloadLibrary)
...
+  ADD_FAILURE() << "Didn't throw exception as expected";
+
```
mikaelarguedas, line 187: unnecessary blank line can be removed

mikaelarguedas:
thanks @cwecht for finding that out!
Code looks good to me except a few cosmetics comments

mikaelarguedas:
Thanks for the update, looks good to me

PR #42, opened 2016-09-20
Add std::unique_ptr API

de-vri-es:
This follows https://github.com/ros/class_loader/pull/38 and adds the API to `pluinglib` as well. That means it requires `indigo-devel` from `class_loader` to build, which is probably not available for the automated tests. I'm not sure what the best solution is for that. It would be nice to release this API addition together with `class_loader`.

Points to note for the review:
- I put the unit test in a separate file. I was tempted to put them in the same file with the other tests, but that would have required compiling those tests with C++11 support as well. I figured it was best not to do that, to make sure no hard requirements on C++11 features sneak in where they shouldn't.
- I switched manual try/catch statements in the unit test with `ASSERT_THROW` for readability. `ASSERT_THROW` checks the exception type, so old behaviour of the test is maintained.
- I also cleaned up some white space errors and long unwrapped comments. I can undo those if desired.


mikaelarguedas:
Thanks for the follow up PR, the change looks good for me.
I'll test them locally on the different distros and release them together to run the tests all at once on the buildfarm.
It's ok to have the cleanup (whitespace/wrapping) commit in the same PR.
Thanks for replicating the c++11 check for compiling the tests.
Will comment here once tested locally

davetcoleman (approved):
I skimmed the changes

mikaelarguedas:
tested locally on both kinetic and indigo without problem

mikaelarguedas (approved):

de-vri-es:
Thanks for the quick review and merge :)

mikaelarguedas:
Released on kinetic and will release on earlier distros once we're sure that no unexpected behaviour emerges

v4hn:
It would be great to have this available in indigo too! So +1 to backport this change to indigo and jade

PR #49, opened 2017-04-16
do not use popen to solve catkin_path.

### LINKED ISSUE #48: getClassLibraryPath() returns abnormal path lists when popen failed with ENOMEM.
Hi.
I found this problem when debugging image_transport library.

To get library path, pluginlib uses popen("catkin_pkg --libs", "r") .
Sometimes popen failes memory is not enough.
On such situation, pluginlib returns incompletion lib paths without any error report. 

I think there are 2 solutions for this problem.
1. throwing exception when popen returns error.
2. instead of popen uses native api.

terakoji:
To fix #48 

Search catkin_path without popen.

### DIFF
--- a/include/pluginlib/class_loader.h
+++ b/include/pluginlib/class_loader.h
@@ -225,10 +225,6 @@ namespace pluginlib
         virtual int unloadLibraryForClass(const std::string& lookup_name);
 
       private:
-        /**
-         * Calls a program from command line and returns output to stdout as a string
-         */
-        std::string callCommandLine(const char* cmd);
 
         /**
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
thanks for the report and the contribution! 
I'll try to review this soon

mikaelarguedas (approved):
thanks @terakoji for the great improvement! the change looks good to me. We could consider exposing this function through the API (not necessarily in pluginlib) because it can be useful to other packages. But we can do that in the future if it becomes necessary

mikaelarguedas:
@ros-pull-request-builder retest this please

PR #60, opened 2017-10-04
remove deprecated macros

wjwwood:
These macros have been deprecated since before Groovy. This pr removes the deprecated macros and the script that replaces them for people.

There are still a few packages which use one of the signatures, `PLUGINLIB_DECLARE_CLASS`, based on a desktop-full code search. I'll make pr's for those.


include/pluginlib/class_list_macros.h
```diff
@@ -39,31 +39,12 @@
...
- */
-#define PLUGINLIB_DECLARE_CLASS(pkg, class_name, class_type, base_class_type) \
-  CLASS_LOADER_REGISTER_CLASS_WITH_MESSAGE(class_type, base_class_type, "pluginlib WARNING: In file " __FILE__ " PLUGINLIB_DECLARE_CLASS is deprecated, please use PLUGINLIB_EXPORT_CLASS instead. You can run the script 'plugin_macro_update' provided with pluginlib in your package source folder to automatically and recursively update legacy macros.  Base = base_class_type, Derived = derived_class_type")
-
-/**
- * @macro This version was only made possible with pluginlib 1.9 series. It's the easiest to use and now the official way of exporting classes.
- * @param - class_type - The real class name with namespace qualifier (e.g. Animals::Lion)
- * @param - base_class_type - The real base class type from which class_type inherits
+ * \param class_type The real class name with namespace qualifier (e.g. Animals::Lion)
+ * \param base_class_type The real base class type from which class_type inherits
```
mikaelarguedas: Nit: I'd rather keep the documentation markup consistent within the repo. We can open a follow up PR to change the syntax in all pluginlib files at once to use the doxygen `\` style

mikaelarguedas (approved):
lgtm, given that this targets melodic I think we can merge it before the follow-up PRs are open. But I can hold this PR if you want to use it as a reminder to update downstream packages relying on these macros

include/pluginlib/class_list_macros.h
wjwwood: Oh, sorry, this slipped in (I have layers of changes 😄). I'll revert this part of this pr.

wjwwood: 7d10fff, but I left the restructuring of the docs because the content changed is on purpose. This should be consistent enough with the existing docs to not stand out I think.

mikaelarguedas: now worries, thanks for the quick fix

PR #61, opened 2017-10-04
remove unused “plugin_tool”

wjwwood:
I'm unable to find any reference to this tool in any documentation, wiki pages, tutorials, or launch files. I believe it was a tool used long ago, but is now unused.

It uses similar arguments to the deprecated pluginlib macros. Some of the utilities it provides might be useful, but I think it would be better integrated into some command line tools in a separate package.

However, some of the functionality is scary, especially `generateAndLoadTypedPluginInterface` which invokes gcc directly to then dlopen a just in time built shared library...

If anyone is using the tool we could restore it pretty easily, but I think we should remove it until someone needs it.


tfoote (approved):
+1 it looks like this was something that Mirza was working on at some point but it never got polished. https://github.com/ros/pluginlib/commits/cf1a7eb5a9a8bb3764c09ce20ac0ab7b5b7e1e4f/src/plugin_tool.cpp

mikaelarguedas (approved):
lgtm, if we can't find any mention of it anywhere I'm in favor of removing it. Thanks!

PR #62, opened 2017-10-04
use a const reference for exception construction

wjwwood:
For example, `std::runtime_error` uses this signature, which is recommended:

http://en.cppreference.com/w/cpp/error/runtime_error


mikaelarguedas (approved):
thanks for the fix

PR #63, opened 2017-10-04
add missing include

wjwwood:
Just noticed this header didn't compile on its own.


mikaelarguedas (approved):
lgtm

mikaelarguedas:
thanks @wjwwood, I forward ported this fix to melodic-devel as well

PR #64, opened 2017-10-10
Style overhaul

wjwwood:
I did this by running `ament_uncrustify` and `ament_cpplint` over the project. The idea of this pr is to get the style of the project close to what we'd like to have even if we cannot easily add automated tests to ensure the style does not diverge. In the future if we can add these tests, the diff will be smaller due to this pr.

I went through and tried to ensure there were no API or ABI changes, but the reviewers should watch closely for these.

I will make another pr to melodic-devel with ABI breaking style changes (using `explicit` on single argument constructors for example).

### DIFF
--- a/include/pluginlib/class_loader.h
+++ b/include/pluginlib/class_loader.h
@@ -47,269 +47,281 @@ namespace pluginlib
 {
 
 #if __cplusplus >= 201103L
-  template<typename T>
-  using UniquePtr = class_loader::ClassLoader::UniquePtr<T>;
+template<typename T>
+using UniquePtr = class_loader::ClassLoader::UniquePtr<T>;
 #endif
+/**
+ * @class ClassLoader
+ * @brief A class to help manage and load classes
+ */
+template<class T>
+class ClassLoader : public ClassLoaderBase
+{
+public:
+  typedef typename std::map<std::string, ClassDesc>::iterator ClassMapIterator;
+
+public:
   /**
-   * @class ClassLoader
-   * @brief A class to help manage and load classes
+   * @brief  Constructor for a ClassLoader
+   * @param package The package containing the base class
+   * @param base_class The type of the base class for classes to be loaded
+   * @param attrib_name The attribute to search for in manifext.xml files, defaults to "plugin"
+   * @param plugin_xml_paths The list of paths of plugin.xml files, defaults to be crawled via ros::package::getPlugins()
+   * @exception pluginlib::ClassLoaderException Thrown if package manifest cannot be found
    */
-  template <class T>
-    class ClassLoader: public ClassLoaderBase
-    {
-      public:
-        typedef typename std::map<std::string, ClassDesc>::iterator ClassMapIterator;
+  ClassLoader(
+    std::string package, std::string base_class,
+    std::string attrib_name = std::string("plugin"),
+    std::vector<std::string> plugin_xml_paths = std::vector<std::string>());
 
-      public:
-        /**
-         * @brief  Constructor for a ClassLoader
-         * @param package The package containing the base class
-         * @param base_class The type of the base class for classes to be loaded
-         * @param attrib_name The attribute to search for in manifext.xml files, defaults to "plugin"
-         * @param plugin_xml_paths The list of paths of plugin.xml files, defaults to be crawled via ros::package::getPlugins()
-         * @exception pluginlib::ClassLoaderException Thrown if package manifest cannot be found
-         */
-        ClassLoader(std::string package, std::string base_class, std::string attrib_name = std::string("plugin"), std::vector<std::string> plugin_xml_paths = std::vector<std::string>());
+  /**
+   * @brief  Destructor for ClassLoader
+   */
+  ~ClassLoader();
 
-        /**
-         * @brief  Destructor for ClassLoader
-         */
-        ~ClassLoader();
+  /**
+   * @brief  Creates an instance of a desired class, optionally loading the associated library automatically if necessary
+   * @param  lookup_name The name of the class to load
+   * @param  auto_load Specifies whether or not to automatically load the library containing the class, set to true by default
+   * @exception pluginlib::LibraryLoadException Thrown when the library associated with the class cannot be loaded
+   * @exception pluginlib::CreateClassException Thrown when the class cannot be instantiated
+   * @return An instance of the class
+   * @deprecated use either createInstance() or createUnmanagedInstance().
+   */
+  __attribute__((deprecated)) T * createClassInstance(
+    const std::string & lookup_name,
+    bool auto_load = true);
 
-        /**
-         * @brief  Creates an instance of a desired class, optionally loading the associated library automatically if necessary
-         * @param  lookup_name The name of the class to load
-         * @param  auto_load Specifies whether or not to automatically load the library containing the class, set to true by default
-         * @exception pluginlib::LibraryLoadException Thrown when the library associated with the class cannot be loaded
-         * @exception pluginlib::CreateClassException Thrown when the class cannot be instantiated
-         * @return An instance of the class
-         * @deprecated use either createInstance() or createUnmanagedInstance().
-         */
-        __attribute__((deprecated)) T* createClassInstance(const std::string& lookup_name, bool auto_load = true);
-
-        /**
-         * @brief  Creates an instance of a desired class (which implicitly calls loadLibraryForClass() to increment the library counter).
-         * Deleting the instance and calling unloadLibraryForClass() is automatically handled by the shared pointer.
-         * @param  lookup_name The name of the class to load
-         * @exception pluginlib::LibraryLoadException Thrown when the library associated with the class cannot be loaded
-         * @exception pluginlib::CreateClassException Thrown when the class cannot be instantiated
-         * @return An instance of the class
-         */
-        boost::shared_ptr<T> createInstance(const std::string& lookup_name);
+  /**
+   * @brief  Creates an instance of a desired class (which implicitly calls loadLibraryForClass() to increment the library counter).
+   * Deleting the instance and calling unloadLibraryForClass() is automatically handled by the shared pointer.
+   * @param  lookup_name The name of the class to load
+   * @exception pluginlib::LibraryLoadException Thrown when the library associated with the class cannot be loaded
+   * @exception pluginlib::CreateClassException Thrown when the class cannot be instantiated
+   * @return An instance of the class
+   */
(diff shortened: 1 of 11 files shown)


include/pluginlib/class_loader.h
```diff
@@ -26,291 +26,337 @@
...
+  T * createUnmanagedInstance(const std::string & lookup_name);
+
+  /// Returns a list of all available plugin manifest paths for this ClassLoader's base class type.
+  /**
+   * @return A vector of strings corresponding to the paths of all available plugin manifests
+   */
+  std::vector<std::string> getPluginXmlPaths();
+
+  /**
+   * @brief  Returns a list of all available classes for this ClassLoader's base class type
```
mikaelarguedas, line 158: any reason for keeping the `@brief Description` rather than using a `/// Description` like for the other functions above ?

include/pluginlib/class_loader.h
```diff
@@ -26,291 +26,337 @@
...
+  /**
+   * @brief  Returns the libraries that are registered and can be loaded
+   * @return A vector of strings corresponding to the names of registered libraries
+   */
+  virtual std::vector<std::string> getRegisteredLibraries();
+
+  /**
+   * @brief Checks if the library for a given class is currently loaded
+   * @param  lookup_name The lookup name of the class to query
+   * @return True if the class is loaded, false otherwise
```
mikaelarguedas, line 220: I know it's not part of the cpplint/uncrustify rules, but it may be good to update the return descriptions to list the valid return values. something like `@return 'true' if the class is loaded, 'false' otherwise` maybe ? (Not sure what's good practice here, I'm looking for feedback not requesting a change)

include/pluginlib/class_loader.h
```diff
@@ -26,291 +26,337 @@
...
+   */
+  bool isClassLoaded(const std::string & lookup_name);
+
+  /**
+   * @brief  Checks if the class associated with a plugin name is available to be loaded
```
mikaelarguedas, line 225: Question: Would it be more consistent to use a single space between keyword and description?

include/pluginlib/class_loader.h
mikaelarguedas: nvm I guess it's because these description fit on a single line so there was no need to switch the format

include/pluginlib/class_loader.h
```diff
@@ -26,291 +26,337 @@
...
+   * @param lookup_name The name of the plugin
+   * @return True if the plugin is available, false otherwise
+   */
+  virtual bool isClassAvailable(const std::string & lookup_name);
+
+  /// Attempt to load the library containing a class with a given name.
```
mikaelarguedas, line 231: Nit: Attempt**s** (or use infinitive in every other description, I don't know which one is best, once decided we should update this PR to adopt it everywhere)

include/pluginlib/class_loader.h
```diff
@@ -26,291 +26,337 @@
...
+   */
+  virtual void loadLibraryForClass(const std::string & lookup_name);
+
+  /**
+   * @brief  Refreshs the list of all available classes for this ClassLoader's base class type
+   * @exception pluginlib::LibraryLoadException Thrown if package manifest cannot be found
+   */
+  virtual void refreshDeclaredClasses();
+
+  /// Decrement the counter for the library containing a class with a given name.
```
mikaelarguedas, line 247: Nit: Decrement**s**

include/pluginlib/class_loader.h
```diff
@@ -26,291 +26,337 @@
...
+   *
+   * @param lookup_name The lookup name of the class to unload
+   * @exception pluginlib::LibraryUnloadException Thrown if the library for the
+   *   class cannot be unloaded
+   * @return The number of pending unloads until the library is removed from memory
+   */
+  virtual int unloadLibraryForClass(const std::string & lookup_name);
+
+private:
+  /// Return the paths to plugin.xml files.
```
mikaelarguedas, line 259: Nit: Return**s**

include/pluginlib/class_loader.h
```diff
@@ -26,291 +26,337 @@
...
+  std::string base_class_;
+  std::string attrib_name_;
+  class_loader::MultiLibraryClassLoader lowlevel_class_loader_;  // The underlying classloader
 };
 
-#include "class_loader_imp.h" //Note: The implementation of the methods is in a separate file for clarity
+}  // namespace pluginlib
+
+// Note: The implementation of the methods is in a separate file for clarity.
+#include "class_loader_imp.h"  // NOLINT
```
mikaelarguedas, line 360: I was originally concerned wit this being now included outside of the pluginlib namespace but it looks like `class_loader_imp.h` defines the pluginlib namespace as well and places not-exposed functions in an anonymous namespace so this should be good as is :+1:

include/pluginlib/class_loader_imp.h
```diff
@@ -74,668 +80,728 @@ std::vector<std::string> catkinFindLib() {
...
 
-      assert(doc_root_node == document.RootElement());
+template<class T>
+std::string ClassLoader<T>::extractPackageNameFromPackageXML(const std::string & package_xml_path)
+/***************************************************************************/
+{
+  tinyxml2::XMLDocument document;
+  document.LoadFile(package_xml_path.c_str());
+  tinyxml2::XMLElement * doc_root_node = document.FirstChildElement("package");
+  if (doc_root_node == NULL) {
```
mikaelarguedas, line 293: Nit/Note to self: not sure how far we want to go down the ros2 style path. we could audit uses of `==` and make sure that the variable is always on the right side of the operator

include/pluginlib/class_loader_imp.h
```diff
@@ -74,668 +80,728 @@ std::vector<std::string> catkinFindLib() {
...
+
+  try {
+    lowlevel_class_loader_.loadLibrary(library_path);
+    it->second.resolved_library_path_ = library_path;
+  } catch (const class_loader::LibraryLoadException & ex) {
+    std::string error_string = "Failed to load library " + library_path +
+      ". Make sure that you are calling the PLUGINLIB_EXPORT_CLASS macro in the "
+      "library code, and that names are consistent between this macro and your XML. "
+      "Error string: "
+      +
```
mikaelarguedas: this looks a bit weird, can we place the `+` on the previous line ?

test/test_base.h
```diff
@@ -1,17 +1,46 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef TEST_BASE_H_
+#define TEST_BASE_H_
```
mikaelarguedas, line 31: Shouldn't the header guard be prepended with the project name ? or is it not the case here because none of this lives in the pluginlib namespace ? (same comment for all test headers)

test/unique_ptr_test.cpp
```diff
@@ -1,81 +1,99 @@
...
-  try
-  {
+  try {
     pluginlib::UniquePtr<test_base::Fubar> foo = test_loader.createUniqueInstance("pluginlib/foo");
     foo->initialize(10.0);
-    EXPECT_EQ(foo->result(),100.0);
-  }
-  catch(pluginlib::PluginlibException& ex)
-  {
+    EXPECT_EQ(foo->result(), 100.0);
```
mikaelarguedas, line 60: Nit/Note to self: expected value first (same applies to all gtests)

test/unique_ptr_test.cpp
```diff
@@ -1,81 +1,99 @@
...
   }
 
-  ROS_INFO( "Checking if plugin is loaded with isClassLoaded..." );
-  if( pl.isClassLoaded( "pluginlib/foo" ) )
-    ROS_INFO( "Class is loaded" );
-  else
-  {
-    FAIL() <<  "Library containing class should be loaded but isn't.";
+  ROS_INFO("Checking if plugin is loaded with isClassLoaded...");
+  if (pl.isClassLoaded("pluginlib/foo") ) {
```
mikaelarguedas: Nit: extra whitespace before closing parenthesis (same below)

mikaelarguedas (commented):
thanks @wjwwood for pounding on this. I did a first round of review with a few comments (mostly questions rather than request to change things).

Some of the comments are unrelated to this code change but can be used as "note to self" to be addressed in the near future if we consider them out of the scope of this PR

include/pluginlib/class_loader.h
wjwwood: Yeah, I only touched the ones that were too long. I'd like to come back and make them all uniform, but I was trying to minimize the changes.

include/pluginlib/class_loader.h
wjwwood: Happy to do that, but I want to do a separate documentation pr where I review each of them.

include/pluginlib/class_loader.h
wjwwood: I think the singular form is preferred, but again I only touched the lines which had to be moved.

I would convert all of them to singular in a documentation fixup pr.

test/test_base.h
wjwwood: I don't know I was just following the suggestion of `ament_cpplint`.

test/unique_ptr_test.cpp
wjwwood: Yeah, I've noticed that uncrustify doesn't catch this always, I briefly investigated it but couldn't find a good setting to fix it.

test/test_base.h
mikaelarguedas: Oh I see, it's because compared to most ROS2 repos, this repo doesnt have a `pluginlib` folder at the root of the repo, so the path from the root of the repository is `test/test_base.h` and not `pluginlib/test/test_base.h` hence the currently suggested header guard. That's something we may consider changing in the future but shouldn't impact this PR

mikaelarguedas (approved):
sounds good to me address all the doc related comments in a follow-up PR :+1: . I'll submit a style (not touching documentation) PR for the "Note to self" ones

PR #66, opened 2017-10-11
create temporary travis file for melodic testing based on lunar

wjwwood:
This is temporary until build.ros.org provides pr testing for melodic.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,34 @@
+# while this doesn't require sudo we don't want to run within a Docker container
+sudo: true
+dist: trusty
+language: python
+python:
+  - "3.5"
+env:
+  global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,34 @@
+# while this doesn't require sudo we don't want to run within a Docker container
+sudo: true
+dist: trusty
```
mikaelarguedas: Don't think it impacts anything given that the job will run in docker but should we make this xenial ?

mikaelarguedas (approved):
thanks for doing this, that'll be very useful for the development of the next few months

mikaelarguedas:
looks like travis did the right thing on the last run, @wjwwood feel free to merge this whenever

PR #67, opened 2017-10-12
Remove and replace boost usage

wjwwood:
This pr removes all uses of boost and replaces them with either c++11 alternatives.

In the case of boost::filesystem, it is replaced with standard library calls to std::filesystem when it is available, or it emulates the subset of std::filesystem required when it is not.

For now this pr includes some of the style commits that are in currently open pr's. I will rebase it when those are merged.

There are still a few things I need to test out before I put this in review, but I'm opening it to get CI going and for visibility.

Also, this removes the version of the function that returns a `boost::shared_ptr` without deprecating it first. I'm not sure if we can do that, but in other places in ROS we've replaced this and just made a typedef for people to use to ease the migration, though it would cause everyone to update their code. I guess it's also possible to retroactively deprecate it in Lunar to give some warning that it will be gone in Melodic, though I'm open to suggestion about how to do that.

### DIFF
--- a/include/pluginlib/impl/filesystem_helper.hpp
+++ b/include/pluginlib/impl/filesystem_helper.hpp
@@ -1,0 +1,186 @@
+/*
+ * Copyright (c) 2017, Open Source Robotics Foundation, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the Willow Garage, Inc. nor the names of its
+ *       contributors may be used to endorse or promote products derived from
+ *       this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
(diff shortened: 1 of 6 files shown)


wjwwood:
There looks like a special so is require to be linked for experimental filesystem to work on Linux. I'll look into it soon.

dirk-thomas:
For a breaking change like this it might be worth posting the idea to discourse for feedback from the community. Also if we break API like this it might make sense to consider doing that in a few packages at once rather then have similar breaking changes over the next few ROS distros.

include/pluginlib/impl/filesystem_helper.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+// The standard library does not provide it, so emulate it.
+#ifndef PLUGINLIB__IMPL__FILESYSYEM_HELPER__HAS_STD_FILESYSTEM
+
+#include <string>
+
+#ifdef _WIN32
+#define CLASS_LOADER_IMPL_OS_DIRSEP '\\'
+#else
+#define CLASS_LOADER_IMPL_OS_DIRSEP '/'
```
mikaelarguedas, line 74: Could filesystem give us the separator (using something like `fs::path::preferred_separator`) allowing us to not have to define them ourselves?

wjwwood: This is the case where std::filesystem is not available...

wjwwood: See: https://github.com/ros/pluginlib/pull/67/files#diff-a438fa6270a9a96b5deb3f4f2835e05cR96

mikaelarguedas: oh sorry, I read it too fast

mikaelarguedas:
> For a breaking change like this it might be worth posting the idea to discourse for feedback from the community. Also if we break API like this it might make sense to consider doing that in a few packages at once rather then have similar breaking changes over the next few ROS distros.

Yes this PR is a proof of concept, we are mostly waiting on the REP3 PR to be out for a more global discussion of allowing getting rid of boost or not in the core packages. I think it will be better to gather feedback rather than package-specific threads on discourse. Based on the outcome of REP3 for melodic, we would either use this only on the ros2 branch, use a `typedef` and update downstream packages accordingly or actually switch to std::shared_ptr everywhere. In the meantime this work is valuable for the ROS2 packages that will need to use pluginlib.

Another alternative if we decide not to change API would be to not get rid of boost only for internal implementation.

@clalancette FYI

wjwwood:
I have a more complete proposal for how to make boost "optional" and to also rename the headers but leave the current name too, but deprecated. It's a bit more scope than "remove boost", so we can either decide to take this and fixup melodic-devel to be API stable (with new deprecations) in follow up pr's, or I can roll this change into a large pr which removes boost and implements the other changes and backward compatibility solutions.

I'll make the "visionary" proposal for moving forward with backwards compatibility in a separate issue and then we can decide what to do with this pr after that.

Either way, this pr is not working at the moment.

wjwwood:
Ok, I've reworked this pr now:

- it's rebased on melodic-devel
- it does not remove the `boost::shared_ptr<T> createInstance(...)` method, but it does deprecate it
- it adds a new method `std::shared_ptr<T> createSharedInstance(...)` meant to replace the boost one

It still replaces the boost filesystem and other functions with c++ alternatives.

I still need to work on linking and usage of the `experimental/filesystem` stuff on Linux.

mikaelarguedas:
@wjwwood I'll do an initial release of pluginlib in melodic without this and will do another release during the alpha period with these changes once rebased and reworked.

@clalancette FYI

wjwwood:
Ok, just a reminder that the filesystem replacement stuff (for systems which don't have it) needs to be updated with fixes from the `ros2` branch.

mikaelarguedas:
> just a reminder that the filesystem replacement stuff (for systems which don't have it) needs to be updated with fixes from the ros2 branch.

:+1: 
The things I have in mind for the "rework" part are the filesystem fixes and the removal of the checking of `__cplusplus` as windows doesn't report c++11 compliance even if it supports it. Do you know of anything else I should keep in mind when reworking this ?

caioaamaral:
How is this pr going?

clalancette:
Given that we don't have another ROS 1 vehicle to ship this in, and given that we have already removed Boost for ROS 2, I'm inclined to close this PR.  @wjwwood do you agree?

wjwwood:
That's fine with me.

PR #70, opened 2017-10-16
New header layout

wjwwood:
This pull request makes the "tick" in a "tick-tock" to rename the headers in this package to have the `.hpp` rather than the `.h` extension. The main goal of this is to minimize the delta between this and the ROS 2 branch, but it also just makes more sense and helps a lot of code editors and style checkers infer the right language.

The headers are renamed (8985e48), the code is updated (97176b3), and then the original header files are restored with forwarding includes and deprecation warnings (dc97479). In the case of the `class_loader_imp.h` header, I just put a `#error` message because no one should be including that directly anyways.

In a future release (n-turtle) we can remove the legacy headers completely.

----

It would be possible to create the new header file names in older releases, forwarding the include to the current header name and without any warning (to allow downstream packages to update to the new include name across ROS distros).

----

If this is taken, then my pr removing boost (https://github.com/ros/pluginlib/pull/67) could be updated to optionally keep the boost function signature when using the old header include, but when including by the new header name. I also plan to update it to use a new function name, deprecating the current function that returns a `boost::shared_ptr`, but I'll add more details about that there.


wjwwood:
btw, `#warning` was the best way to deprecate a file that I could find, if anyone has a better way to do it, let me know.

mikaelarguedas:
> btw, `#warning` was the best way to deprecate a file that I could find, if anyone has a better way to do it, let me know.

This seem to be a common practice that I find better than the pretty popular alternative using a `#pragma message` :+1: so I'm totally fine with it except if someone suggests a "better" way of doing it

include/pluginlib/class_loader_imp.h
```diff
@@ -1,806 +1,36 @@
...
-template<class T>
-int ClassLoader<T>::unloadClassLibraryInternal(const std::string & library_path)
-/***************************************************************************/
-{
-  return lowlevel_class_loader_.unloadLibrary(library_path);
-}
-
-}  // namespace pluginlib
+#error Header <pluginlib/class_loader_imp.h> is deprecated, \
+but this file should not be included directly, include <pluginlib/class_loader.hpp> instead.
```
mikaelarguedas: cpplint being unable to figure out that this is a message. It complains about the use of `not` here. We could either replace it with "never" or silence cpplint on this line

include/pluginlib/class_desc.h
```diff
@@ -1,84 +1,38 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
 
 #ifndef PLUGINLIB__CLASS_DESC_H_
 #define PLUGINLIB__CLASS_DESC_H_
 
-#include <string>
+#warning Including header <pluginlib/class_desc.h> is deprecated, \
+include <pluginlib/class_desc.hpp> instead.
```
mikaelarguedas: uncrustify will complain about these not being indented (same everywhere below)

include/pluginlib/class_desc.h
```diff
@@ -1,84 +1,38 @@
...
-*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
-*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
-*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
-*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-*  POSSIBILITY OF SUCH DAMAGE.
-*
-*********************************************************************/
+/*
+ * Copyright (c) 2017, Open Source Robotics Foundation, Inc.
```
mikaelarguedas: our current copyright/licence checker will need the licence to be explicitly stated to match the licence template. we would need to add `Software License Agreement (BSD License 2.0)` between the `All rights reserved` line and and the body of the licence for it to work. (Assuming that `ament_copyright` eventually gets updated to handle lines with multiples leading whitespaces and block comments)

test/test_plugins.cpp
```diff
@@ -27,7 +27,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <pluginlib/class_list_macros.h>
+#include <pluginlib/class_list_macros.hpp>
 #include "test_base.h"  // NOLINT
 #include "test_plugins.h"  // NOLINT
```
mikaelarguedas, line 32: Should these include statements also be updated to use `#include "./HEADER_NAME"` rather than being explicitly ignored by cpplint?

mikaelarguedas (commented):
Looks good, a few nitpicks/style comments inline

---
> It would be possible to create the new header file names in older releases, forwarding the include to the current header name and without any warning (to allow downstream packages to update to the new include name across ROS distros).

Yeah I think that's a good idea so that people can share code more easily across distros. I'll likely backport this to the `kinetic-devel` branch, I am not sure yet what changes I'll propagate to `indigo-devel` given that I haven't looked into backporting the other modifications we merged recently.

---

> I just put a #error message because no one should be including that directly anyways.

Arguably if any package breaks because of this change it is because they are including an internal implementation header when they should not, so I think it would be ok to propagate this change as well.

---
Thanks for renaming pluginlib_exceptions to exceptions but keeping the name of the original header :+1:

include/pluginlib/class_desc.h
wjwwood: I was gonna leave that until we are able to turn on the copyright linter for the melodic-devel branch (if we're ever able to do so). These files won't exist on the ros2 branch.

test/test_plugins.cpp
wjwwood: Maybe, I'd say out-of-scope for this pr.

include/pluginlib/class_desc.h
mikaelarguedas: Given that the files original files (with the Willow Garage copyright) have the licence listed explicitly and the files is ROS 2 packages have it as well, I would add it to these new files to stay consistent.
Then that won't be enough for it to pass the linter and I agree with you, we can address cosmetics later when/if we can enable the copyright linter

test/test_plugins.cpp
mikaelarguedas: fair enough :+1:

include/pluginlib/class_desc.h
wjwwood: I copied this from this file:

https://github.com/ros/pluginlib/blob/9e79885c2b4997e7b4abb3a3a607b5e098e0f469/include/pluginlib/class_loader.h#L1-L28

Also, these are new files, so that's why I didn't include the Willow Garage copyright.

What exactly do you want me to do here? I'm not sure I follow what you mean about explicitly listing the license.

mikaelarguedas: Oh looks like it was inconsistent in the first place, so that's fine, we can merge thisas is :+1:  and will iterate if we end up turning on the linter.

---
What I meant (for completeness) is that this file originally started with 
```
* Software License Agreement (BSD License)
 *
 *  Copyright (c) 2008, Willow Garage, Inc.
 *  All rights reserved.
```

and now has only:
```
+ * Copyright (c) 2017, Open Source Robotics Foundation, Inc.
+ * All rights reserved.
```
and it looks like the linter expects this additional line as part of the BSD header (https://github.com/ament/ament_lint/blob/767257ff141d7957f3c8c7b92106aced95794537/ament_copyright/ament_copyright/template/bsd2_header.txt)

wjwwood: I see what you mean now. I added the explicit name in 7aeabc9, we'll still have to work on making this consistent when we turn on the linter.

mikaelarguedas:
Sounds good, thanks for iterating, would you mind squashing the follow up commits so that we can merge this PR as 3 commits? (renaming, updating, adding backward compatible headers)

wjwwood:
Done.

mikaelarguedas (approved):
Thanks :+1:

v4hn:
:boom: *WHY* :boom:
This breaks so much code for no real benefit....

It's to late for complains now, anyway... :unamused:

mikaelarguedas:
@v4hn It's never too late :)

The main intent is to allow us to keep the ROS 1 and ROS 2 branches as close as possible, compliant with the ROS 2 style, and to allow to eventually provide a C interface to pluginlib.

As we didn't have time to do all the refactor we wanted to do in pluginlib on time for melodic, I agree with you this doesn't have much benefit currently except making our life easier on the ROS2 side. 
I tried to provide an easy migration path by releasing migration scripts as well as opening PRs so that people don't have to migrate themselves.
While this doesn't "break" any code for the time being, I understand that the newly introduced warnings are inconvenient. I am considering removing the deprecation warnings in melodic and adding them back in a future ROS version when the refactor of pluginlib is more advanced and that the API changes will justify this small extension change.
Either ways I will keep providing both versions as well as scripts to perform the migration so that people that want to change can easily do it.

But we would like to get rid of them when we reach the point that this diff is the only code diff between the ROS 1 and ROS 2 version of this package.

I'm happy to consider additional feedback, so please feel free to keep commenting here.

wjwwood:
+1 to just removing the warning in the `.h` headers for Melodic.

Maintainers who want to migrate now can do so, and others do not have to yet.

And as you said, we can reconsider removing the `.h` version of the headers in the future via a tick-tock deprecation if it becomes the only patch between the ROS 1 and ROS 2 branches.

v4hn:
The main problem I see is that you assume all software is well-maintained.
Sadly that's not true in our line of work.
So with each API change you introduce, you make life harder for researchers
to try out software of previous generations...
After all one good thing about the core ros libraries is that their APIs are rather stable, but still supported.

+1 to removing the warning until it is crucial for ROS1-2 support.

wjwwood:
Never changing isn’t an option either though. So I’m in favor of avoiding a required change for this case, but thinking nothing will ever change in any future ros distrubtions is unreasonable in my opinion. Also this change doesn’t break anything, at worst it will cause a warning. In the next ros release it would break but there are other more substantive changes that may occur in the future.

v4hn:
There is a difference between "no changes" and "no changes without acceptable justification".

It's not like I have never done such changes myself with only disputable justifications.
Nevertheless, I think "force users to provide ROS2 (style) compatibility"
is no acceptable reason to remove a header (after deprecating it now).

> but there are other more substantive changes that may occur in the future.

Once users have to edit their code anyway because of required API changes,
I would not complain about changing the header name.

PR #88, opened 2017-11-16
Continue loading classes on error (#85)

mikaelarguedas:
* continue loading classes on error

* construct string with file rather than adding new API

* match style of the rest of the file

* missing whitespace

@wjwwood FYI


wjwwood (approved):
👍, thanks for the heads up.

PR #92, opened 2017-12-13
Windows contains older version of std filesystem (still experimental)

bfjelds:


mikaelarguedas:
Thanks @bfjelds for the contribution.

I have 2 questions regarding filesystem on Windows:
- The MSDN [filesystem page](https://msdn.microsoft.com/en-us/library/hh874694.aspx) shows a v1 namespace under the filesystem namespace: `using namespace std::experimental::filesystem::v1;`. Do you know if this is still necessary (we're targeting VS2015 update 3 and MSVC 19.0.24215.1 at the moment) ?
- Is including `<filesystem>` necessary of is `<experimental/filesystem>` enough ? I see that we're including only the latter in [other places](https://github.com/ros2/demos/blob/fc0705beb7ce297da9221938f402fb2691287861/composition/src/api_composition.cpp#L35-L38) of the codebase that we will need to update accordingly if Windows requires to include both. 

I'll let @wjwwood review this as well as he's been the one implementing this filesystem helper.

bfjelds:
These are great questions...I was focused on getting pluginlib to build 8n the context of implementing laser_filters, but was using VS2017.

I'll investigate further in about a week and see if I can't get some definitive answers :)

bfjelds:
The v1 is an inline namespace (new to me!) ... if we wanted to refer specifically to v1, we could, but it isn't required.

<experimental/filesystem> is a wrapper around <filesystem>at this point, so either would work.  I went with experimental/filesystem to be consistent with the namespace.

This was all verified on VS2015 Update 3 and VS2017.

include/pluginlib/impl/filesystem_helper.hpp
```diff
@@ -35,7 +35,19 @@
 #ifndef PLUGINLIB__IMPL__FILESYSTEM_HELPER_HPP_
 #define PLUGINLIB__IMPL__FILESYSTEM_HELPER_HPP_
 
-#if defined(__has_include)
+#if defined(_MSC_VER)
```
mikaelarguedas, line 38: I think it would make sense to compare the value to a minimal version of compiler as some Windows platforms with an older compiler will have the variable defined but not the filesystem header. I didnt track down the first version of MSVC implementing it but as we target VS2015 Update3 we should be safe by requiring `_MSC_VER >= 1900`

mikaelarguedas (commented):
Thanks @bfjelds for checking this.

The code looks good to me, I have one comment regarding the MSC_VER check.

After looking a bit into it, it seems that we need this special case for MSVC only because VS2015 doesn't support `__has_include` whereas [VS2017](https://blogs.msdn.microsoft.com/vcblog/2017/08/11/c17-features-and-stl-fixes-in-vs-2017-15-3/), [gcc](https://gcc.gnu.org/projects/cxx-status.html) and [clang](https://clang.llvm.org/cxx_status.html) implement the feature. I think it failed to build for you with VS2017 beccause both `<filesystem>` and `<experimental/filesystem>` are present, and as we check the presence of `<filesystem>` first you ended up with `namespace fs = std::filesystem;` while the Windows implementation is under the namespace `std::experimental::filesystem`.

mikaelarguedas:
I rebased this branch and added the change I suggested regarding the version check on this branch https://github.com/ros/pluginlib/tree/ros_ros2_mikael.
@bfjelds would you mind confirming that this will allow you to build your plugins ? If so I'll push the changes to this branch and merge it

bfjelds:
Confirmed that it builds...looks great, thanks!!

mikaelarguedas:
Thanks @bfjelds for testing it!
I opened #96 with the suggested change on top of the current ros2 branch to avoid us to have to resolve the conflicts manually. Thanks again for reporting this and submitting a fix.

Note to self: this should be closed once #96 merged

wjwwood:
@mikaelarguedas it would be nice to keep @bfjelds's commits as the foundation of the pr, to preserve his attribution, in #92 you can't tell that he had a hand in it from the commits. I'll try to fix up on of these pr's so that it is preserved.

wjwwood:
Closing in favor of https://github.com/ros/pluginlib/pull/96, for some reason I could edit this pr on the website, but I couldn't push to your branch @bfjelds.

PR #94, opened 2017-12-15
Ros2 reenable tests

mikaelarguedas:
This is work in progress.
Opening this for visibility as @wjwwood may have been working on this before the release.
Not sure if the test should be move to a different package for both cleaner separation and work around the ament index issues (of pluginlib not finding itself in the index at test time)


wjwwood:
> Not sure if the test should be move to a different package for both cleaner separation and work around the ament index issues (of pluginlib not finding itself in the index at test time)

This is why I basically avoided doing this at the time. I was waiting on pluginlib to be in a subfolder so I could make a `test_pluginlib` that worked through the normal ament_index channels.

mikaelarguedas:
> This is why I basically avoided doing this at the time. I was waiting on pluginlib to be in a subfolder so I could make a test_pluginlib that worked through the normal ament_index channels.

:+1: Then I'll bring this closer to melodic-devel that already has the subfolder and will reenable the tests after that. thanks for the feedback

mjcarroll:
Replaced by #130

PR #97, opened 2018-01-23
adjust library search to work on windows, warn about lib prefix

wjwwood:
This is required for rviz to load `.dll`'s on Windows which reside in `bin`. Also, all users are probably including the implicit `lib` prefix for their libraries explicitly, e.g.:

https://github.com/ros2/rviz/blob/4e10a6925d8db08e539ea54997abeb1ed6226767/rviz_default_plugins/plugins_description.xml#L1

This is not so portable because the cmake code uses `<library_name>` which generates a library like `lib<library_name>.so` on Linux and `lib<library_name>.dylib` on macOS, but `<library_name>.dll` on Windows.

So, I added the ability for pluginlib to search with and without the `lib` prefix, i.e. if you give `libfoo` it will also search for `foo`, and if you give `foo` it will also search for `libfoo`. However, giving slight preference to the user input without the `lib` prefix, I added a warning that prints if you give `libfoo` which suggests you use `foo` instead.

Long term, I think we should deprecate this searching and instead use compile time information to store the actual location (relative or absolute) of the library and simply match that with what is in the plugin description `.xml` by name.

### DIFF
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -376,11 +376,13 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
   //
   //   - <prefix for rviz>/lib/librviz_default_plugins.so
   //   - <prefix for rviz>/lib64/librviz_default_plugins.so
+  //   - <prefix for rviz>/bin/rviz_default_plugins.dll
   //   - <prefix for rviz>/lib/rviz/librviz_default_plugins.so
   //   - <prefix for rviz>/lib64/rviz/librviz_default_plugins.so
   //
   // The extension, e.g. `.so`, might be different based on the operating
   // system, e.g. it might be `.dylib` on macOS or `.dll` on Windows.
+  // Similarly, the library might have the `lib` prefix added or removed.
   // Also, the library name might have a `d` added if the library is built
   // debug, depending on the system.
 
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Thanks @wjwwood, I'll have a look soon

wjwwood:
Here's ROS 2 CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3892/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1011/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3218/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3988/)

pluginlib/include/pluginlib/class_loader_imp.hpp
```diff
@@ -390,39 +392,77 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
...
     package_prefix + path_separator + "lib",
     package_prefix + path_separator + "lib64",
+    package_prefix + path_separator + "bin",  // also look in bin, for dll's on Windows
     package_prefix + path_separator + "lib" + path_separator + exporting_package_name,
     package_prefix + path_separator + "lib64" + path_separator + exporting_package_name,
+    package_prefix + path_separator + "bin" + path_separator + exporting_package_name,
   };
+
+  // Prepare to setup the relative file paths.
   bool debug_library_suffix = (0 == class_loader::systemLibrarySuffix().compare(0, 1, "d"));
```
mikaelarguedas, line 411: Note that currently `systemLibrarySuffix` will [always return the non debug version](https://github.com/ros/class_loader/commit/68796036d722808817b1ad54f7cb6600aa07107d) (but that's great to keep logic for handling both)

pluginlib/include/pluginlib/class_loader_imp.hpp
```diff
@@ -390,39 +392,77 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
...
-  std::string stripped_library_name_with_extension = stripped_library_name + non_debug_suffix;
 
-  for (const auto & current_path : all_paths_without_extension) {
-    all_paths.push_back(current_path + path_separator + library_name_with_extension);
-    all_paths.push_back(current_path + path_separator + stripped_library_name_with_extension);
+  std::string library_name_alternative;  // either lib<library> or <library> without lib prefix
+  const char * lib_prefix = "lib";
+  if (library_name.rfind(lib_prefix, 0) == 0) {
+    library_name_alternative = library_name.substr(strlen(lib_prefix));
+    RCUTILS_LOG_WARN_NAMED("pluginlib.ClassLoader",
```
mikaelarguedas, line 424: :+1: thanks for adding the warning that'll help converge to the new way of defining them

mikaelarguedas (approved):
Lgtm, thanks ! :+1:

wjwwood:
Thanks for the quick review @mikaelarguedas 👍

PR #99, opened 2018-02-09
use new class_loader header names

mikaelarguedas:
related to ros/class_loader#80

PR #101, opened 2018-03-03
use new class_loader errors and fix deprecation warnings

mikaelarguedas:
redo of #99 applied to melodic-devel


mikaelarguedas:
failure on bionic expected as all ROS packages depending on console_bridge and built before the last bionic package sync from debian stable are not installable anymore. Merging as this passes without warnings on debian stretch

PR #115, opened 2018-04-15
remove deprecation warning until sers are required to change code

mikaelarguedas:
based on discussion from https://github.com/ros/pluginlib/pull/70#issuecomment-379435585. We will remove deprecation warnings until we roll out more refactor to pluginlib at wich point we will deprecate them gain and remove them in the following distribution.


mikaelarguedas:
Ignoring travis failure as there's no travis job anymore

PR #116, opened 2018-05-14
Fix compile error with gcc 8.1.0

jerry73204:
Gcc 8.1.0 fails to compile filesystem_helper.hpp with -std=c++14. The cause is that it passes the check "__has_include(<filesystem>)". However, std::filesystem is not defined. We add an additional check "__cplusplus >= 201703L".

### DIFF
--- a/pluginlib/include/pluginlib/impl/filesystem_helper.hpp
+++ b/pluginlib/include/pluginlib/impl/filesystem_helper.hpp
@@ -49,7 +49,7 @@ namespace fs = std::experimental::filesystem;
 # define PLUGINLIB__IMPL__FILESYSYEM_HELPER__HAS_STD_FILESYSTEM
 
 #elif defined(__has_include)
-# if __has_include(<filesystem>)
+# if __has_include(<filesystem>) && __cplusplus >= 201703L
 #  include <filesystem>
 
 namespace pluginlib
--- a/pluginlib/include/pluginlib/impl/filesystem_helper.hpp
+++ b/pluginlib/include/pluginlib/impl/filesystem_helper.hpp
@@ -36,8 +36,9 @@
 #define PLUGINLIB__IMPL__FILESYSTEM_HELPER_HPP_
 
 
-#if _MSC_VER >= 1900
-# include <experimental/filesystem>
+#if defined(_MSC_VER)
+# if _MSC_VER >= 1900
+#  include <experimental/filesystem>
 namespace pluginlib
 {
 namespace impl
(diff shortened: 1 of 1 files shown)


mjbogusz:
I think better option would be to use `--std=c++17` (or `gnu++17`).

The issue comes from GCC 8 providing `<filesystem.h>` in addition to `<experimental/filesystem.h>`, however the first one provides `std::filesystem` only on C++17 and newer standards (and does not provide `std::experimental::filesystem` at all) and the latter uses old, experimental implementation.

This solution would however require switching to C++17 for all of ROS2 (or at least for all packages using pluginlib directly or nondirectly).

jerry73204:
I agree. It requires some work to migrate whole ros2 to C++17. The fix here is intended to be a workaround before the we switch to -std=c++17, and makes sure it compiles with newer GCC 8.

mikaelarguedas:
The change looks sensible to me.
It should not impact MSVC (that tends to not set `__cplusplus` properly) as it is evaluated before.
Clang seems happy with the change as well: [Build Status](http://ci.ros2.org/job/ci_osx/3602/).

@jerry73204 Were you able to build the entire ROS2 stack with this change ?

Running CI on the entire stack (gcc5.4 for Linux):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4408/) (false positive cmake warning)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1358/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3603/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4465/)

jerry73204:
I followed the instructions on [Arch wiki](https://wiki.archlinux.org/index.php/ROS#Ros_2) to build ros2. @mikaelarguedas is right. I would do some surveys to make proper fix.

jerry73204:
I slightly changed the statements to avoid incorrect __cplusplus value with MSVC. For those non-MSVC compilers, we test the headers with `__has_include`. Though `__has_include` is adopted as standard since C++17, it presents with -std=c++{11,14} when compiling with g++ >= 5 or clang++. So I assume it work under most circumstances.

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4411/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1359/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3604/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4467/)

mikaelarguedas (approved):
lgtm

mikaelarguedas:
@wjwwood do you mind having a quick :eyes: when you have a chance ?

wjwwood (approved):
It’s fine by me.

mikaelarguedas:
thanks @jerry73204 for the patch!

PR #123, opened 2018-07-19
Fix build issue when build on Windows

johnsonshih:
Minor changes to make it build on Windows

### DIFF
--- a/pluginlib/include/pluginlib/class_loader.hpp
+++ b/pluginlib/include/pluginlib/class_loader.hpp
@@ -90,7 +90,7 @@ public:
    * \return An instance of the class
    * \deprecated use either createInstance() or createUnmanagedInstance()
    */
-  ROS_DEPRECATED T * createClassInstance(
+  [[deprecated]] T * createClassInstance(
     const std::string & lookup_name,
     bool auto_load = true);
 
--- a/pluginlib/include/pluginlib/class_loader.hpp
+++ b/pluginlib/include/pluginlib/class_loader.hpp
@@ -90,7 +90,7 @@ public:
    * \return An instance of the class
    * \deprecated use either createInstance() or createUnmanagedInstance()
    */
-  __attribute__((deprecated)) T * createClassInstance(
+  ROS_DEPRECATED T * createClassInstance(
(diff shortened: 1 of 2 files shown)


pluginlib/include/pluginlib/class_loader.hpp
```diff
@@ -90,7 +90,7 @@ class ClassLoader : public ClassLoaderBase
    * \return An instance of the class
    * \deprecated use either createInstance() or createUnmanagedInstance()
    */
-  __attribute__((deprecated)) T * createClassInstance(
+  ROS_DEPRECATED T * createClassInstance(
```
mikaelarguedas: would `[[deprecated]]` work for your use case? this would allow to stay closer to the `ros2` branch

pluginlib/include/pluginlib/class_loader_imp.hpp
```diff
@@ -547,7 +547,11 @@ std::string ClassLoader<T>::getPathSeparator()
 /***************************************************************************/
 {
 #if BOOST_FILESYSTEM_VERSION >= 3
+#ifdef _WIN32
```
mikaelarguedas: can you please indent these for readability?
```
#if foo
#  if bar
...
#  endif
#else
...
#endif
```

pluginlib/include/pluginlib/class_loader.hpp
johnsonshih: the compiler that I use support[[deprecated]] but I'm not sure if other compilers all support C++14.  If all compilers used to build ROS support it, I can change it to [[deprecated]], please advise.

mikaelarguedas: ROS Melodic assumes C++14 compatible compilers (see [REP3](http://www.ros.org/reps/rep-0003.html#melodic)), so feel free to go ahead and switch to `[[deprecated]]`

johnsonshih: cool, change to use standard attribute

mikaelarguedas (approved):
lgtm, thanks for the contribution and the iteration!

PR #124, opened 2018-08-02
Fix build issue on Windows, add linking library for test_plugins

johnsonshih:


mikaelarguedas:
@johnsonshih do you have the exact error message when trying to build the code as is ?

The current change seems to be overlinking as the test doesnt use any boost symbols. Can you check if #125 fixes the issue you are facing ?

johnsonshih:
Yes, #125 is a better change, please use that one, I'll close this PR.

PR #125, opened 2018-08-06
link test_plugins against class_loader for windows compilation

mikaelarguedas:
`test_plugins` needs to be able to register classes with class_loader so needs to link against it.

Replaces #124


mikaelarguedas:
Thanks @johnsonshih for the original report and patch!

PR #132, opened 2018-12-03
adding ament_package() to test_pluginlib

mlautman:
Without this change sourcing your environment gives you the following output:

```
not found: "/home/mike/ws_ros2/install/test_pluginlib/share/test_pluginlib/local_setup.sh"
not found: "/home/mike/ws_ros2/install/test_pluginlib/share/test_pluginlib/local_setup.bash"
```

### DIFF
--- a/test_pluginlib/CMakeLists.txt
+++ b/test_pluginlib/CMakeLists.txt
@@ -37,3 +37,5 @@ if(BUILD_TESTING)
     target_compile_definitions(${PROJECT_NAME}_utest PUBLIC "PLUGINLIB__DISABLE_BOOST_FUNCTIONS")
   endif()
 endif()
+
+ament_package()


mjcarroll (approved):

dirk-thomas:
While this patch fixes the symptom it should be changed. The `test_pluginlib` only exists for unit tests and shouldn't install anything - e.g. not have a Debian package and not appear in `ros2 pkg list`.

nuclearsandwich:
These packages are properly ignored by bloom and not released into Crystal debs https://github.com/ros2-gbp/pluginlib-release/blob/dae90d5d995b7db85c79c60c3e3e92a9f3319865/crystal.ignored

dirk-thomas:
> These packages are properly ignored by bloom and not released into Crystal debs

Correct, I was mainly describing the desired end state. The problem still exists in from-source builds.

PR #130, opened 2018-10-25
Ros2 reorganize tests

mjcarroll:
This will replace Mikael's rework in #94.

This has the addition of a `test_pluginlib_fixture` package to exercise the full extent of `pluginlib`.

PR #134, opened 2018-12-10
Collapse testing back into pluginlib package.

mjcarroll:
Merges testing back into the pluginlib package.

Depends on ros2/tinyxml2_vendor#3


mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5846/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2410/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4842/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5744/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5847/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2411/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4843/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5745/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5848/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2412/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4844/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5746/)

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5850/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2414/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4846/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5748/)

pluginlib/CMakeLists.txt
```diff
@@ -30,47 +30,96 @@ install(
...
+  target_compile_definitions(test_plugins PUBLIC "PLUGINLIB__DISABLE_BOOST_FUNCTIONS")
+  target_compile_definitions(test_plugins PRIVATE "TEST_PLUGINLIB_FIXTURE_BUILDING_LIBRARY")
+  target_include_directories(test_plugins
+      PUBLIC
+      include
+      test/include
+      ${ament_index_cpp_INCLUDE_DIRS}
+      ${class_loader_INCLUDE_DIRS}
+      ${rcutils_INCLUDE_DIRS}
+      ${TinyXML2_INCLUDE_DIRS}
```
dirk-thomas: For ROS 2 packages you could use `ament_target_dependencies` to add include directories, libraries etc. to each target.

pluginlib/test/package.xml
```diff
@@ -1,14 +1,18 @@
 <?xml version="1.0"?>
```
dirk-thomas: Assuming this manifest file is only used for testing and to find the plugin I would suggest to remove most of its content: dummy name, dummy version, specific description, no authors, dependencies, etc.

pluginlib/CMakeLists.txt
```diff
@@ -30,47 +30,96 @@ install(
...
+    )
+
+  # The $<$<CONFIG:Debug>:> prevents MSVC from making Release/Debug dirs.
```
dirk-thomas: Does this work for all build types?

nuclearsandwich (commented):
I tested this on Xenial and the pluginlib tests ran without issue.

I am somewhat concerned that the fixture package.xml could be picked up by some tools (not colcon, clearly as the build did not fail) as part of a valid package all its own. Since the file is copied out during the test phase this could be mitigated by naming the source file something like fixture_package.xml instead.

I liked the acceptance tests when they were introduced as a top-level package in https://github.com/ros/pluginlib/pull/130 but I'm not sure if it's worth the CMake hoops here to have them.

dirk-thomas:
> I am somewhat concerned that the fixture package.xml could be picked up by some tools

The spec forbids nesting of packages so that shouldn't be a problem with compliant tools.

nuclearsandwich:
> > 
> 
> The spec forbids nesting of packages so that shouldn't be a problem with compliant tools.

No more `find -name package.xml` for me :sob:

mjcarroll:
I will do as you suggest and change the name, so it won't be picked up by simple searches in the source space, at least, but it's good to know that it shouldn't choke any of the tooling.

mjcarroll:
Nevermind on renaming, because CMake doesn't support it on `file(COPY` or `file(INSTALL`, it will remain package.xml.

New CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5861/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2417/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4849/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5751/)

dirk-thomas:
> CMake doesn't support it on `file(COPY` or `file(INSTALL`, it will remain package.xml.

`configure_file (... COPY_ONLY)` might be an option.

pluginlib/CMakeLists.txt
mjcarroll: Tested on Windows and Linux with `CMAKE_BUILD_TYPE=Debug` and `CMAKE_BUILD_TYPE=Release`

My understanding is that the generator function should always evaluate empty, regardless of the build configuration.  I added some documentation above it as well as where I got it from.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5864/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2420/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4852/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5754/)

mjcarroll:
Just realized I had some vestigial debug output that needed to be removed.

nuclearsandwich (approved):
:+1: :bowing_man:

nuclearsandwich:
This needs ros2/tinyxml2_vendor#3 before merging

dirk-thomas (approved):

pluginlib/CMakeLists.txt
```diff
@@ -30,47 +30,85 @@ install(
...
+  )
+  if(TARGET ${PROJECT_NAME}_utest)
+    ament_target_dependencies(
+      ${PROJECT_NAME}_utest
+      class_loader
+      ament_index_cpp
+      rcutils
+      TinyXML2
+    )
+    target_link_libraries(${PROJECT_NAME}_utest ${TinyXML2_LIBRARIES})
```
sloretz, line 103: Nit, I think `ament_target_dependencies(... TinyXML2)` above does this already.

https://github.com/ament/ament_cmake/blob/2d65d41e047e203cd0531df9b0987d8aaea7d6ca/ament_cmake_target_dependencies/cmake/ament_target_dependencies.cmake#L55-L56

mjcarroll: Correct, this was an accident.  Thanks!

PR #138, opened 2019-01-02
Fix dev build dep on TinyXML2

mjcarroll:
Tests weren't properly depending on tinyxml2_vendor.

### DIFF
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -86,6 +86,7 @@ if(BUILD_TESTING)
       class_loader
       ament_index_cpp
       rcutils
+      tinyxml2_vendor
       TinyXML2
     )
     target_compile_definitions(${PROJECT_NAME}_unique_ptr_test PUBLIC "PLUGINLIB__DISABLE_BOOST_FUNCTIONS")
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Was this causing specific errors somewhere?

mjcarroll:
Yes, on Windows, if I recall correctly, if you didn't depend on tinyxml2_vendor, then TinyXML2 wouldn't be found.

nuclearsandwich:
@mjcarroll it seems like we haven't actually needed this change for our Windows builds nor have I seen any user reports about it.

At first I was hoping to avoid explicit dependencies on our vendor packages but without this patch we're basically just "getting lucky" with TinyXML2 being available prior to finding tinyxml2_vendor. So let's get CI sorted and get this merged.

nuclearsandwich:
>  So let's get CI sorted and get this merged.

We dropped Travis from the ROS 2 branch in #142. The Travis CI failure is invalid and can be ignored.

mjcarroll:
Do you want me to rebase?

nuclearsandwich:
> Do you want me to rebase?

The PR will get squash-merged and there aren't any current conflicts. As long as we use the branch feature on ci.ros2.org this branch will be merged into the ros2 one for CI. But it can't hurt.

PR #140, opened 2019-01-12
strip 'lib' prefix when loading library on Windows

kejxu:
Windows does not add `lib` prefix to libary name, if 
```
std::string library_path = getClassLibraryPathFromLibraryName(library_name, it->second.package_);
```
fails to find the library path, remove the prefix and try again

this code change does not alter existing logic on Linux

### DIFF
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -16,6 +16,11 @@ if(CATKIN_ENABLE_TESTING)
 
   add_library(test_plugins EXCLUDE_FROM_ALL SHARED test/test_plugins.cpp)
   target_link_libraries(test_plugins ${class_loader_LIBRARIES})
+  if(WIN32)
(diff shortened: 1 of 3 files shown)


seanyen:
cc @nuclearsandwich

nuclearsandwich:
@ros-pull-request-builder retest this please.

nuclearsandwich:
@seanyen, @kejxu  because this PR adds the new function to a public interface `getClassLibraryPathFromLibraryName` it breaks C++ ABI compatibility which we don't wish to do for melodic.

Our options then become 

1. Targeting the change at noetic-devel only.
2. Targeting the change at noetic-devel and creating a backportable version which doesn't add to the public interface.
3. Refactoring the PR not to affect the public interface and targeting melodic-devel (which will be incorporated into noetic-devel).

I'm tempted toward 3 mostly because I don't see a strong utility in the split functions but I am open to input.

seanyen:
@nuclearsandwich Thanks for the feedback! I will be looking into options to not break the ABI.

PR #141, opened 2019-01-12
 set Dll destination to CATKIN_PACKAGE_LIB_DESTINATION

kejxu:
currently [catkin logic on Windows](https://github.com/ros/catkin/blob/kinetic-devel/cmake/platform/windows.cmake) would publish dll's to global binary destination for devel space:
```
RUNTIME_OUTPUT_DIRECTORY ${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_BIN_DESTINATION}
```

however, this is mainly needed for ROS core packages for backward compatibility support. For the vast majority of packages, publishing to the package's bin directory would be okay.

for `test_plugin`, given it's a plugin, its look-up path is hard-coded in [test_plugins.xml](https://github.com/ros/pluginlib/blob/melodic-devel/pluginlib/test/test_plugins.xml#L1) as `lib\...`, so it has to be exported to `CATKIN_PACKAGE_LIB_DESTINATION`

### DIFF
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -16,6 +16,11 @@ if(CATKIN_ENABLE_TESTING)
 
   add_library(test_plugins EXCLUDE_FROM_ALL SHARED test/test_plugins.cpp)
   target_link_libraries(test_plugins ${class_loader_LIBRARIES})
+  if(WIN32)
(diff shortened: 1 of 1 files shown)


kejxu:
this change should become unnecessary since the behavior could get covered by https://github.com/ros/pluginlib/pull/145

hopefully that one will get merged soon! =)

PR #144, opened 2019-02-08
Add stdc++fs as a target link library for clang compiler on linux

bhatsach:
This change is needed to compile pluginlib package and packages which
depend on it on linux clang compiler


bhatsach:
### Points to note
* I am adding a nightly job on ROS2 CI to build on linux clang compiler in order to detect bugs early. Mac OS X builds can take very long to complete. Here's the PR link for the same: [ros2/ci#244](https://github.com/ros2/ci/pull/244)

### Testing done
* I have tested the above changes by doing a colcon build locally with all the ros2 packages pulled

tfoote:
This approach seems like it's trying to fix the test and not to identify the more general underlying requirement. What is the differentiating factor that requires this flag being added? Can you enumerate the cases, because at a glance at this looks like it should just be detecting linux an adding the additional compile flag? The previous flag was effectively selecting linux as osx runs clang by default and windows is vsbuild.

bhatsach:
> This approach seems like it's trying to fix the test and not to identify the more general underlying requirement. What is the differentiating factor that requires this flag being added? Can you enumerate the cases, because at a glance at this looks like it should just be detecting linux an adding the additional compile flag? The previous flag was effectively selecting linux as osx runs clang by default and windows is vsbuild.

Makes sense, I may be able to just update the check to verify that we are on linux. Will look into this now.

bhatsach:
Updated the PR with just a check for linux for linking experimental filesystem libraries. Also updated the check to linux while exporting the libraries. Experimental Filesystem is a linux specific concept rather than compiler specific.

bhatsach:
@nuclearsandwich and @tfoote, please let me know if you have any comments on the latest revision, thanks

pluginlib/CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ if(BUILD_TESTING)
     )
     target_compile_definitions(${PROJECT_NAME}_unique_ptr_test PUBLIC "PLUGINLIB__DISABLE_BOOST_FUNCTIONS")
 
-    if(CMAKE_COMPILER_IS_GNUCXX)
+    if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
```
nuclearsandwich: It seems like most of the CMAKE pundits on StackOverflow and mailing lists favor the construct

```cmake
 if (UNIX AND NOT APPLE)
  ...
endif()
```

While it's a bit obtuse I think it's likely what's desired in order to also cover cases like NetBSD, QNX or NuttX.

@wjwwood I would love your advice on whether you think explicitly matching Linux or using this other construct is more appropriate.

nuclearsandwich (commented):
One question about the CMake logic for long term *nix support. Once that is settled we'll need to run CI.

pluginlib/CMakeLists.txt
wjwwood: Not sure if it makes much difference, since it will break one way or another. The only thing I can see that might be better about `if (UNIX AND NOT APPLE)` is that it will break earlier (in pluginlib tests) rather than later (in a package like rviz).

nuclearsandwich: > Not sure if it makes much difference, since it will break one way or another.

Not sure I fully understand. What's expected to break? Is it the availability of the c++fs on non-Linux platforms?

wjwwood: For example, NuttX may not have `c++fs` (not sure if this is true or not).

nuclearsandwich: Okay. That makes sense to me. If the platform doesn't have the required dependency then the build won't succeed one way or the other. But if the dependency is satisfied on a platform the `(UNIX AND NOT APPLE)` query should implement the desired link.

 I'm all in favor of failing earlier and being forward compatible. So let's go with `(UNIX AND NOT APPLE)`.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6152/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2620/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5074/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6015/)
* Linux Clang [Build Status](https://ci.ros2.org/job/ci_linux/6153/)

nuclearsandwich (approved):
Will merge as long as CI comes back green.

bhatsach:
> One question about the CMake logic for long term *nix support. Once that is settled we'll need to run CI.

Thanks a lot for identifying and updating this. I locally tested it on ubuntu and the change looks good :)

nuclearsandwich:
@bhatsach do you have any insight into the warnings that are coming back on the Linux Clang job?

nuclearsandwich:
Because the warnings are in other packages and don't seem directly related to pluglinlib and the supported builds have no errors. I'm going to merge with the expectation that those interested in getting the clang on Linux builds fully green will do so.

PR #145, opened 2019-02-20
also search under bin if on Windows

kejxu:
search under bin if on Windows

similar logics has already been added/implemented in the [`ros2` branch](https://github.com/ros/pluginlib/blob/ros2/pluginlib/include/pluginlib/class_loader_imp.hpp#L416), this could be considered as a back-port as well as a fix for Windows


seanyen:
@nuclearsandwich Just a friendly reminder for #140 #145 in case they are out of your radar.

seanyen:
@nuclearsandwich @dirk-thomas sorry to ping you again on this pull request. Now the [ROS on Windows](https://aka.ms/ros/buildfarm) has switched to consume all the upstream repositories for `ros_base`, except for this one. And I would love to see how we can move #140 and #145 forward and migrate to use the upstream `pluginlib` as well. Thanks!

nuclearsandwich (approved):

PR #150, opened 2019-04-11
class_loader_imp: do not call virtual function in destructor

cwecht:
This fixes clang-tidys "clang-analyzer-optin.cplusplus.VirtualCall"-warning.

Also this makes more clear, which function is actually called.

### DIFF
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -100,7 +100,7 @@ ClassLoader<T>::~ClassLoader()
 /***************************************************************************/
 {
   ROS_DEBUG_NAMED("pluginlib.ClassLoader", "Destroying ClassLoader, base = %s, address = %p",
-    getBaseClassType().c_str(), this);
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Humbly begging your forgiveness @cwecht for the tardy review. I discussed this change with @clalancette and here is what we observed:
* The current destructor could result in a crash when the virtual function is called since the implementation is in the subclass whose destructor has already been run.
* With this patch, any subclass which overloads `getBaseClassType` will not have its implementation used here.

With those observations in mind, averting a crash is highly desirable but the tradeoff is that anyone out there overriding getBaseClassType won't have their implementation used by the debug message.

We discussed making getBaseClassType private and non-virtual but that only works if no one is actually using overrides.

sloretz:
Thank you for the PR!

Unfortunately I don't think we should merge this one. Given the concerns about behavior changes and that [ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160), I think we should leave the behavior as is.

PR #154, opened 2019-05-07
Export tinyxml2 libraries downstream

esteve:
This change makes the tinxyml2 libraries available to downstream packages when they are crosscompiled. The calls to `find_package` are in the extras file because the tinyxml2 libraries may be installed system-wide and therefore have absolute paths, which can't be resolved when crosscompiling.

### DIFF
--- a/pluginlib/pluginlib-extras.cmake
+++ b/pluginlib/pluginlib-extras.cmake
@@ -35,3 +35,7 @@ if(UNIX AND NOT APPLE)
   # ament_export_libraries() because it is not absolute and cannot be found with find_library
   list(APPEND pluginlib_LIBRARIES ${FILESYSTEM_LIB})
 endif()
+
+find_package(tinyxml2_vendor REQUIRED)
+find_package(TinyXML2 REQUIRED)
+list(APPEND pluginlib_LIBRARIES ${TinyXML2_LIBRARIES})


Karsten1987:
@nuclearsandwich Are we considering this for Dashing?

nuclearsandwich:
If I had been more attentive to it it could likely have been included but I failed to take action earlier. I will add it to the first Dashing patch release queue.

esteve:
@Karsten1987 @nuclearsandwich is there anything I should addressing in this PR? Thanks.

pluginlib/pluginlib-extras.cmake
```diff
@@ -35,3 +35,7 @@ if(UNIX AND NOT APPLE)
   # ament_export_libraries() because it is not absolute and cannot be found with find_library
   list(APPEND pluginlib_LIBRARIES ${FILESYSTEM_LIB})
 endif()
+
+find_package(tinyxml2_vendor REQUIRED)
+find_package(TinyXML2 REQUIRED)
+list(APPEND pluginlib_LIBRARIES ${TinyXML2_LIBRARIES})
```
nuclearsandwich, line 41: Have we just been getting lucky that pluginlib_LIBRARIES hadn't needed the TinyXML2_LIBRARIES added before because they were being brought in from somewhere else?

pluginlib/pluginlib-extras.cmake
```diff
@@ -35,3 +35,7 @@ if(UNIX AND NOT APPLE)
   # ament_export_libraries() because it is not absolute and cannot be found with find_library
   list(APPEND pluginlib_LIBRARIES ${FILESYSTEM_LIB})
 endif()
+
+find_package(tinyxml2_vendor REQUIRED)
```
nuclearsandwich, line 39: I'm somewhat hesitant about finding the vendor package here as someday we won't need one for tinyxml2 and tinyxml2_vendor when used at install time should register CMake configuration that locates TinyXML2 for builds using the same install space... but maybe that doesn't work for isolated builds when tinyxml2_vendor is in the same workspace.

It's probably fine as-is but I'm curious if removing tinyxml2_vendor still allows successful cross-compilation with tinyxml2_vendor in an underlay and in the same workspace as pluginlib.

nuclearsandwich:
>  is there anything I should addressing in this PR? Thanks.

Nope I've just had a devil of a time clearing my review backlog.

I created an `export-tinyxml2` branch on this fork and merged these changes into it for CI.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7541/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3678/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6176/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7368/)

nuclearsandwich (approved):
I asked a couple of clarifying questions but this should be good to merge as-is as long as there are no surprises during CI.

esteve:
@nuclearsandwich sorry for the slow response, I somehow missed this, I'll reply to your questions.

pluginlib/pluginlib-extras.cmake
esteve: It was fine the way it was, however when crosscompiling the paths in `${TinyXML2_LIBRARIES}` where absolute in the host, so the downstream packages wouldn't find them when they were crosscompiled for the target and libraries/executables didn't have all the symbols. One way to know if crosscompiled builds will be fine is to add the following as linker flags:

```cmake
if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-undefined")
elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-undefined,error")
endif()
```

this will fail the builds even if they are not crosscompiled by making sure that there are no missing symbols.

esteve: That's what we added to https://gitlab.com/autowarefoundation/autoware.ai/common/blob/master/autoware_build_flags/cmake/autoware_build_flags-extras.cmake#L1 This code could be added to `ament_cmake_ros`, for example.

pluginlib/pluginlib-extras.cmake
esteve: I recall testing these changes without `tinyxml2_vendor` and not working, but it's been a while.

nuclearsandwich: I think it makes sense for it not to work if tinyxml2_vendor is in the same workspace but if TinyXML2 is available from the base system or in an underlay workspace via tinyxml2_vendor it should work without it. But that's not a condition worth testing for and we'll have to pull the vendor packages out by hand if we retire them no matter what.

PR #171, opened 2020-01-26
[noetic][pluginlib] declare specific boost dependencies

mikaelarguedas:
May be worth targeting to a noetic-devel to not impact released packages on other distributions

related to https://github.com/ros/ros_comm/pull/1871

PR #174, opened 2020-02-14
Update CMake minimum version to address CMP0048 warning.

nuclearsandwich:
Hat tip to Shane whose comment in another PR diagnosed the warning:
https://github.com/ros/pluginlib/pull/171#issuecomment-586474749

Melodic's currently supported platforms, Stretch and Bionic, use CMake
3.7.2 and 3.10.2 respectively so this change can target melodic-devel.


nuclearsandwich:
Closing in favor of #173

PR #176, opened 2020-02-25
[Windows] strip 'lib' prefix when loading library.

seanyen:
This is a follow up of #140.

Many ROS packages are originally targeted for Linux and there is a convention to put the library prefix `lib` when describing `<library path="lib/libplugin">`. However, on Windows, the shared library doesn't have such naming convention. As a result, `class loader` is not able to find the library by names.

This pull request is to propose that `class loader` should also search the library name without `lib` prefix on Windows, which is as a mitigation to help downstream packages don't need to modify their plugin XML but still can load the plugin binaries.

This pull request is also fix an little problem that `stripAllButFileFromPath()` is returning the path separator in the result string. Based on the context of `getAllLibraryPathsToTry()`, the path separator seems to be an extra.

### DIFF
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -357,12 +357,30 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
   std::string stripped_library_name = stripAllButFileFromPath(library_name);
   std::string stripped_library_name_with_extension = stripped_library_name + non_debug_suffix;
 
+#ifdef _WIN32
+  // try to strip the leading `lib` from the library path.
+  // Windows doesn't have the `lib` prefix convention for shared libraries.
+  const std::string lib_suffix = "lib";
+  std::string stripped_library_name_win32 = stripped_library_name;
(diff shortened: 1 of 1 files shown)


seanyen:
@nuclearsandwich This is a rework for #140. And this time I only touch up `getAllLibraryPathsToTry()` and hope this won't break any interfaces.

seanyen:
@nuclearsandwich Friendly ping. And any feedback is appreciated. I hope this change won't be slipped for the next `pluginlib` release window. :)

seanyen:
@nuclearsandwich @sloretz Following up. Any feedback would be appreciated.

pluginlib/include/pluginlib/class_loader_imp.hpp
```diff
@@ -805,7 +826,7 @@ std::string ClassLoader<T>::stripAllButFileFromPath(const std::string & path)
   if (std::string::npos == c) {
     return path;
   } else {
-    return path.substr(c, path.size());
```
sloretz, line 808: What does this change do?

sloretz:
ROS 1 is end-of-life (EOL) as May 31st 2025, so I'm closing this.

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #177, opened 2020-02-28
fix getCatkinLibraryPaths on 64bit

vmatare:
Not every distribution has this symlink from `lib` to `lib64` that Ubuntu has on 64bit systems. The portable, safe bet is to use `lib64` on a 64bit system, and `lib` on a 32bit system. This issue exists in other places in ROS as well, but it was most prominent in the pluginlib. Long term, the check macros should be moved to a more central location, but I think for the time being, this PR will ease many people's lives ;-)

### DIFF
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -51,6 +51,7 @@
 #include "boost/bind.hpp"
 #include "boost/filesystem.hpp"
 #include "boost/foreach.hpp"
+#include "boost/predef/architecture.h"
 #include "class_loader/class_loader.hpp"
 
 #include "ros/package.h"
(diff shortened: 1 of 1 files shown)


sloretz:
Thank you for the PR!

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160). Every change comes with a risk of introducing regressions, and there isn't much time left to fix them. To make sure this PR doesn't introduce any regressions please:

* Describe how you tested this change
* Recruit at least one more person to review this PR and try it out on their system

sloretz:
ROS 1 is end-of-life (EOL) as May 31st 2025, so I'm closing this.

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #183, opened 2020-03-24
Fix filesystem linking on clang9

emersonknapp:
The nightly Linux builds are now running on Focal which has LLVM 9, which no longer provides filesystem as a separate library and instead provides it as part of libc++

Intended to resolve red status on https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/

Signed-off-by: Emerson Knapp <emerson.b.knapp@gmail.com>

### DIFF
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -33,6 +33,8 @@ install(
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
 
+  # NOTE this block is also defined in pluginlib-extras.cmake
+  # changes here need to be duplicated there
   if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_FLAGS MATCHES "-stdlib=libc\\+\\+")
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0)
       # Before LLVM 7.0, filesystem is part of experimental
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -35,9 +35,14 @@ if(BUILD_TESTING)
 
   if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_FLAGS MATCHES "-stdlib=libc\\+\\+")
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0)
+      # Filesystem introduced to experimental in LLVM 7.0
       set(FILESYSTEM_LIB c++experimental)
-    else()
+    elseif(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
+      # Before LLVM 9.0 you have to manually link the fs library
       set(FILESYSTEM_LIB c++fs)
+    else()
+      # Starting at LLVM 9.0 filesystem is built in
+      set(FILESYSTEM_LIB)
     endif()
   else()
     set(FILESYSTEM_LIB stdc++fs)
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -35,7 +35,7 @@ if(BUILD_TESTING)
 
   if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_FLAGS MATCHES "-stdlib=libc\\+\\+")
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0)
-      # Filesystem introduced to experimental in LLVM 7.0
+      # Before LLVM 7.0, filesystem is part of experimental
       set(FILESYSTEM_LIB c++experimental)
     elseif(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
       # Before LLVM 9.0 you have to manually link the fs library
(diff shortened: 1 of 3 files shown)


emersonknapp:
@emersonknapp - please run this CI job
Gist: https://gist.githubusercontent.com/emersonknapp/697c20b3ae1d0fc1c49b45a7d1312010/raw/0c95531d6cf4f2fda613ef8204fb3d04719414a6/ros2.repos
BUILD args:
TEST args: --packages-select pluginlib
Job: ci_launcher nightly_linux_clang_libcxx

piraka9011 (approved):
LGTM on Green CI. Build seems to be failing for a separate reason.
Does this provide any significant optimization tho?

emersonknapp:
This isn't an optimization, just a build fix for a new version of the compiler

emersonknapp:
* Nightly linux clang libcxx [Build Status](https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/364/)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9774/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5393/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7942/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/9639/)

thomas-moulard (approved):

zmichaels11 (approved):

emersonknapp:
* Nightly Linux Clang libcxx [Build **Status](https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/365/)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9779/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5400/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7945/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/9643/)

nuclearsandwich:
@emersonknapp did you have a reason for not scoping your CI builds to `--packages-up-to pluginlib`?

emersonknapp:
Since the cmake extras affect the build of dependent packages it needed to be everything upstream as well. Maybe `--packages-up-to pluginlib --packages-above pluginlib` would be more directed

nuclearsandwich:
> Maybe `--packages-up-to pluginlib --packages-above pluginlib` would be more directed

The package selection arguments take the intersection so that would just build pluginlib. You'd need to separately list both sets first and use `--packages-select` with it.

dirk-thomas:
> Since the cmake extras affect the build of dependent packages it needed to be everything upstream as well.

Packages depending on `pluginlib` are called downstream packages.

emersonknapp:
Hence just build everything, no `packages-through` option... the slow stuff is like ogre vendor, and rviz uses pluginlib so probably not much to be gained

emersonknapp:
> Packages depending on pluginlib are called downstream packages.

You're right, but you also got the point :)

nuclearsandwich:
Looking at the rosbag2_storage build failure in Linux/clang, [this header's include check for std::filesystem and >=C++17](https://github.com/ros/pluginlib/blob/37e1763edb988e3c4e07de76a4e22163a81b5219/pluginlib/include/pluginlib/impl/filesystem_helper.hpp#L39-L56) is coming back negative resulting in it finding and including experimental::filesystem and triggering the deprecation warning.

If we don't make any changes to that header, I think the downstream packages all need to be moved to C++17. Bumping the C++ version in rosbag2_storage to C++17 resolved that build error locally. The other option to explore is silencing that deprecation warning as we currently do on Windows.

dirk-thomas:
Foxy won't be requiring C++17 since we currently don't use any C++17 features. So it needs stay compatible with C++14 compilers.

nuclearsandwich:
In that case @emersonknapp and I will need to look closer and figure out why this change resolves the issue for pluginlib's own build but not rosbag2_storage's on top of it.

nuclearsandwich:
After looking at combinations of C++14, 17 and experimental/filesystem, filesystem locally I'm not sure why this PR in its current state doesn't trigger the deprecation error during pluginlib's own test compilation. But I am pretty sure that the right answer to using C++14 with experimental/filesystem via clang10 is to suppress the deprecation. 

@emersonknapp do you want to pull https://github.com/ros/pluginlib/commit/79576967292a7955c42f433cbb1ccb1802cdb9a9 into your branch and re-run the full CI gamut? I ran [Build Status](https://ci.ros2.org/job/ci_linux/9825/) which just includes pluginlib and rosbag2_storage. This should be enough to get us building on clang to address the warnings.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9827/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5450/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7991/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9688/)
* Nightly linux clang libcxx [Build Status](https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/373/)

The OSX warnings are from https://github.com/ros2/rclcpp/pull/1034.  The clang libcxx build also has the rclcpp warning (seems to be a clang thing), otherwise it's all coming from vendor packages - console_bridge, poco, spdlog, cyclonedds, Fast-RTPS.

emersonknapp:
@nuclearsandwich I think we're good to merge with this, see analysis in previous comment for yellow builds. Thanks for the extra commit.

nuclearsandwich (approved):
Thanks for contributing!

rotu:
@nuclearsandwich @dirk-thomas @emersonknapp  Even if we're targeting C++14, the correct change *is* to set CMAKE_CXX_STANDARD to 17, as explained here https://github.com/ros/pluginlib/pull/182#issuecomment-606383370. By silencing the warning, you will break the build.

PR #184, opened 2020-03-31
Bump to c++17

rotu:
replaces #182

This allows us to use std::filesystem, as the logic in filesystem_helper sniffs the current c++ standard. CMake should fall back gracefully to c++14 if c++17 is not available.

This is a correct solution but not the best one, which would be add a library which wraps filesystem and which we link to instead of imposing compilation constraints on consuming projects

### DIFF
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -1,9 +1,8 @@
 cmake_minimum_required(VERSION 3.5)
 project(pluginlib)
 
-# Default to C++14
-if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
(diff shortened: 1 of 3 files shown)


rotu:
Discussion here: https://github.com/ros/pluginlib/pull/182
Note:

> _even if_ we're targeting CMake 14, setting `CMAKE_CXX_STANDARD` to a newer version will fall back to [the previous supported standard](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD) if 17 is unavailable.

> - `experimental/filesystem` is not part of either C++14 nor C++17, and in a compiler that implements C++17 should not be used.
> - `std::filesystem` is not part of C++14.

clalancette:
Obsoleted by #212, so closing.

PR #186, opened 2020-04-08
Using rcpputils for library names

ahcorde:
I have seen this error on the buildfarm in PluginLib. Requesting for a missing reference `class_loader::systemLibrarySuffix` this function was removed from class_loader. Using rcpputils to get

```
(.text._ZN9pluginlib11ClassLoaderIN9test_base5FubarEE23getAllLibraryPathsToTryERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESB_[_ZN9pluginlib11ClassLoaderIN9test_base5FubarEE23getAllLibraryPathsToTryERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESB_]+0x415): undefined reference to `class_loader::systemLibrarySuffix[abi:cxx11]()'
21:35:03 /usr/bin/ld: unique_ptr_test.cpp:
```

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -463,7 +463,7 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
         all_paths.push_back(current_search_path + path_separator + current_library_path);
       }
     }
-  } catch (const std::runtime_error& ex) {
+  } catch (const std::runtime_error & ex) {
     throw std::runtime_error{ex.what()};
   }
 
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -458,7 +458,6 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
       for (auto && current_library_path : all_relative_library_paths) {
         all_paths.push_back(current_search_path + path_separator + current_library_path);
       }
-      // We're in debug mode, try debug libraries as well
       for (auto && current_library_path : all_relative_debug_library_paths) {
         all_paths.push_back(current_search_path + path_separator + current_library_path);
       }
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -68,6 +68,7 @@
 #include "ament_index_cpp/get_resource.hpp"
 #include "ament_index_cpp/get_resources.hpp"
 #include "class_loader/class_loader.hpp"
+#include "rcpputils/shared_library.hpp"
 #include "rcutils/logging_macros.h"
 
 #include "./class_loader.hpp"
@@ -423,14 +424,6 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
     package_prefix + path_separator + "bin" + path_separator + exporting_package_name,
   };
 
-  // Prepare to setup the relative file paths.
-  bool debug_library_suffix = (0 == class_loader::systemLibrarySuffix().compare(0, 1, "d"));
-  std::string non_debug_suffix;
-  if (debug_library_suffix) {
-    non_debug_suffix = class_loader::systemLibrarySuffix().substr(1);
-  } else {
-    non_debug_suffix = class_loader::systemLibrarySuffix();
-  }
   std::string stripped_library_name = stripAllButFileFromPath(library_name);
 
   std::string library_name_alternative;  // either lib<library> or <library> without lib prefix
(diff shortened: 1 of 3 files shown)


emersonknapp (approved):
LGTM, let's get this in soon, all builds that depend on pluginlib are now broken

pluginlib/include/pluginlib/class_loader_imp.hpp
```diff
@@ -446,30 +439,32 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
...
+    for (auto && current_search_path : all_search_paths) {
+      for (auto && current_library_path : all_relative_library_paths) {
+        all_paths.push_back(current_search_path + path_separator + current_library_path);
+      }
+      // We're in debug mode, try debug libraries as well
       for (auto && current_library_path : all_relative_debug_library_paths) {
         all_paths.push_back(current_search_path + path_separator + current_library_path);
       }
     }
+  } catch (const std::runtime_error& ex) {
```
Karsten1987:
```suggestion
  } catch (const std::runtime_error & ex) {
```

pluginlib/include/pluginlib/class_loader_imp.hpp
```diff
@@ -446,30 +439,32 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
...
+      rcpputils::get_platform_library_name(library_name_alternative, true),
+      rcpputils::get_platform_library_name(stripped_library_name, true),
+      rcpputils::get_platform_library_name(stripped_library_name_alternative, true)
+    };
+
+    for (auto && current_search_path : all_search_paths) {
+      for (auto && current_library_path : all_relative_library_paths) {
+        all_paths.push_back(current_search_path + path_separator + current_library_path);
+      }
+      // We're in debug mode, try debug libraries as well
```
Karsten1987: That `if` clause has gone. How do we know that we're in debug mode? Or do we return always both, debug and non-debug and the comment could simply go?

emersonknapp:
Note this depends on https://github.com/ros2/rcpputils/pull/52 and https://github.com/ros2/rcutils/pull/227

pluginlib/include/pluginlib/class_loader_imp.hpp
```diff
@@ -446,30 +439,32 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
...
+    std::vector<std::string> all_relative_debug_library_paths = {
+      rcpputils::get_platform_library_name(library_name, true),
+      rcpputils::get_platform_library_name(library_name_alternative, true),
+      rcpputils::get_platform_library_name(stripped_library_name, true),
+      rcpputils::get_platform_library_name(stripped_library_name_alternative, true)
+    };
+
+    for (auto && current_search_path : all_search_paths) {
+      for (auto && current_library_path : all_relative_library_paths) {
+        all_paths.push_back(current_search_path + path_separator + current_library_path);
```
Karsten1987, line 459: nitpick: `all_paths` could be pre-allocated with `reserve` (`all_search_paths.size() * all_relative_library_paths * 2`)

pluginlib/include/pluginlib/class_loader_imp.hpp
nuclearsandwich: Karsten beat me to raising this point. Is the intent to always search debug paths as well?

pluginlib/include/pluginlib/class_loader_imp.hpp
```diff
@@ -423,14 +424,6 @@ std::vector<std::string> ClassLoader<T>::getAllLibraryPathsToTry(
     package_prefix + path_separator + "bin" + path_separator + exporting_package_name,
   };
 
-  // Prepare to setup the relative file paths.
```
nuclearsandwich, line 426: Can you please add to the commit message, PR description, or both information about the removal of this block and either where the code was moved or why it can be removed.

ahcorde: I removed the `debug_mode` variable because poco was returning the suffix in case we are in debug mode or not. I was trying to fix this fast. I will dig more

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10008/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5609/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8144/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9897/)

pluginlib/include/pluginlib/class_loader_imp.hpp
nuclearsandwich: If https://github.com/ros/pluginlib/pull/186/files#diff-83eae4c2eef591307fdaf5e9d11dc2c1L427 was always true previously, that's sufficient reasoning to drop the conditional. While it'd be interesting to know why I think it's academic.

ahcorde: can I do somethin like ?

```
bool debug_mode;
#if defined(_DEBUG)
	debug_mode = true;
#else
	debug_mode = false;
#endif
```

emersonknapp:
I'd like to bring up the point that we should have reverted https://github.com/ros/class_loader/pull/139 before working on these PRs. Both https://github.com/ros2/rcpputils/pull/52 and https://github.com/ros2/rcutils/pull/227 were under-CI'd (they only built and tested up to the package at being changed, when they are widely used). 

The fixes should be fine, but I think we need to be holding a much more revert-friendly engineering culture in the ROS 2 core. I think we need to do an unqualified CI build when changing core utilities.

Before merging _this_ pr, I think we need a CI run that builds all of ros2.repos, not just `--packages-up-to pluginlib` <-- that doesn't prove that this fixes the issue downstream

Karsten1987:
I agree with @emersonknapp. Can we easily revert the commit?

as for CI, I think it has to be compiled in full and test all packages above pluginlib

dirk-thomas:
> Before merging this pr, I think we need a CI run that builds all of ros2.repos, not just `--packages-up-to pluginlib` <-- that doesn't prove that this fixes the issue downstream

Since currently builds on top of `pluginlib` are certainly broken this fix (if deemed fine by reviewers) should be merged right away for the fastest path to unbreaking builds. Waiting for full CI builds for 2+h doesn't help (assuming the offending changes aren't reverted right away).

emersonknapp:
Yes, I agree we're already too far down this hole, but I'm trying to make the point that we shouldn't have merged the rcpputils and rcutils fixes before reverting the original offending change, for future policy

nuclearsandwich (approved):
:heavy_check_mark: Introduction of rcpputils dependency.
:heavy_check_mark: No direct poco dependency to remove as it was being used via class_loader, which is still being used.
:heavy_check_mark: Some context has been added to the PR to justify removing the debug check and conditional here.
:black_square_button: Green CI. I would recommend also running builds in Debug configuration and perhaps builds that build packages downstream of pluginlib

ahcorde:
This job will run over the full ros2.repos and it will test test_rclcpp and test_communication

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10009/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5610/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8145/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9899/)

pluginlib/include/pluginlib/class_loader_imp.hpp
nuclearsandwich: I don't know what poco was doing behind the scenes which the previous check was piggybacking on. It might be worth introducing such a check into rc{,pp}utils in order to properly replace our usage of it.

PR #189, opened 2020-04-27
add missing export of stdc++fs and TinyXML2 via modern CMake

dirk-thomas:
Related to ros2/ros2#904.

### DIFF
--- a/pluginlib/pluginlib-extras.cmake
+++ b/pluginlib/pluginlib-extras.cmake
@@ -44,3 +44,10 @@ endif()
 find_package(tinyxml2_vendor REQUIRED)
 find_package(TinyXML2 REQUIRED)
 list(APPEND pluginlib_LIBRARIES ${TinyXML2_LIBRARIES})
+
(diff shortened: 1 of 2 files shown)


pluginlib/pluginlib-extras.cmake
```diff
@@ -44,3 +44,10 @@ endif()
 find_package(tinyxml2_vendor REQUIRED)
 find_package(TinyXML2 REQUIRED)
 list(APPEND pluginlib_LIBRARIES ${TinyXML2_LIBRARIES})
```
hidmic, line 46: @dirk-thomas do we have to export `tinyxml2` libraries along with `pluginlib`'s now that it is an exported interface?

dirk-thomas: Yes, the modern CMake approach is only added - it doesn't remove the classic CMake variable to maintain backward compatibility.

hidmic (approved):

PR #190, opened 2020-05-05
link against tinyxml2 correctly

### LINKED ISSUE #191: [ros2] tinyxml2_vendor not in export set
When trying to export targets which depend on pluginlib, CMake complains that `tinyxml2_vendor` is not part of any export set.
I've seen this independent from each other within rosbag2 and image_transport.

```
Starting >>> image_transport
--- stderr: image_transport
CMake Error: install(EXPORT "image_transport" ...) includes target "image_transport" which requires target "tinyxml2_vendor" that is not in any export set.
CMake Generate step failed.  Build files cannot be regenerated correctly.
make: *** [cmake_check_build_system] Error 1
---
Failed   <<< image_transport	[ Exited with code 2 ]
```

Does it make sense to export the vendor package in the first place? I mean, it makes sense for pluginlib to use the vendor package to link correctly to `tinyxml2`, but I am not seeing any reason ad-hoc to also export the vendor package if the `pluginlib` target transitively links correctly against `tinyxml2`.

Karsten1987:
closes #191 
Connects to ros2/rosbag2#402

I don't know exactly why but it looks to me as if `tinyxml2_vendor` as well as `TinyXML2` are not correctly linked to the interface target of pluginlib.

By explicitly linking to `tinyx2ml::tinyxml2` - and not adding the vendor package to the targets - the error message in https://github.com/ros2/rosbag2/pull/403#issuecomment-623623630 is being resolved.

As a general question, why should pluginlib export a dependency on the vendor package and not just directly transitively link to tinyxml2?

### DIFF
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -21,8 +21,9 @@ find_package(TinyXML2 REQUIRED)  # provided by tinyxml2 upstream, or tinyxml2_ve
 add_library(${PROJECT_NAME} INTERFACE)
 target_include_directories(${PROJECT_NAME} INTERFACE
   "$<INSTALL_INTERFACE:include>")
+target_link_libraries(${PROJECT_NAME} INTERFACE tinyxml2::tinyxml2)
 ament_target_dependencies(${PROJECT_NAME} INTERFACE
-  ament_index_cpp class_loader rcutils rcpputils tinyxml2_vendor TinyXML2)
+  ament_index_cpp class_loader rcutils rcpputils)
 ament_export_dependencies(ament_index_cpp class_loader rcutils rcpputils tinyxml2_vendor TinyXML2)
 ament_export_include_directories(include)
 ament_export_targets(${PROJECT_NAME})
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -21,9 +21,8 @@ find_package(TinyXML2 REQUIRED)  # provided by tinyxml2 upstream, or tinyxml2_ve
 add_library(${PROJECT_NAME} INTERFACE)
 target_include_directories(${PROJECT_NAME} INTERFACE
   "$<INSTALL_INTERFACE:include>")
-target_link_libraries(${PROJECT_NAME} INTERFACE tinyxml2::tinyxml2)
 ament_target_dependencies(${PROJECT_NAME} INTERFACE
-  ament_index_cpp class_loader rcutils rcpputils)
+  ament_index_cpp class_loader rcutils rcpputils TinyXML2)
 ament_export_dependencies(ament_index_cpp class_loader rcutils rcpputils tinyxml2_vendor TinyXML2)
 ament_export_include_directories(include)
 ament_export_targets(${PROJECT_NAME})
--- a/pluginlib/pluginlib-extras.cmake
+++ b/pluginlib/pluginlib-extras.cmake
@@ -40,14 +40,3 @@ if(UNIX AND NOT APPLE)
   # ament_export_libraries() because it is not absolute and cannot be found with find_library
   list(APPEND pluginlib_LIBRARIES ${FILESYSTEM_LIB})
 endif()
-
-find_package(tinyxml2_vendor REQUIRED)
-find_package(TinyXML2 REQUIRED)
-list(APPEND pluginlib_LIBRARIES ${TinyXML2_LIBRARIES})
-
-add_library(tinyxml2_vendor INTERFACE IMPORTED)
-target_include_directories(tinyxml2_vendor INTERFACE
-  ${TinyXML2_INCLUDE_DIRS})
-target_link_libraries(tinyxml2_vendor INTERFACE
-  ${TinyXML2_LIBRARIES})
-#list(APPEND pluginlib_TARGETS tinyxml2_vendor)
--- a/pluginlib/pluginlib-extras.cmake
+++ b/pluginlib/pluginlib-extras.cmake
@@ -50,4 +50,4 @@ target_include_directories(tinyxml2_vendor INTERFACE
   ${TinyXML2_INCLUDE_DIRS})
 target_link_libraries(tinyxml2_vendor INTERFACE
   ${TinyXML2_LIBRARIES})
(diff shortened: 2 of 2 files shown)


Karsten1987:
@dirk-thomas do you mind having a look at this? Not sure exactly how to move forward.

dirk-thomas:
Since the `else` case in the CMake module provided by the vendor package doesn't have an interface target `tinyx2ml::tinyxml2` (see https://github.com/ros2/tinyxml2_vendor/blob/c89337439ffe9b5123ca310af0d746eb9cbfa110/cmake/Modules/FindTinyXML2.cmake#L180) I doubt the current approach works in all the cases.

Karsten1987:
That makes total sense.

I think it would make sense to adapt that `else` logic in the vendor package with the following to make sure that there is a _modern_ cmake target available:
```
  add_library(tinyxml2 STATIC IMPORTED)
  set_property(TARGET tinyxml2 PROPERTY IMPORTED_LOCATION ${TinyXML2_LIBRARIES})
  set_property(TARGET tinyxml2 PROPERTY INCLUDE_DIRECTORIES ${TinyXML2_INCLUDE_DIRS})
  list(APPEND tinyxml2_vendor_TARGET tinyxml2)
```

Alternatively, this code snippet could be placed inside the pluginlib's cmake logic and guarded with `if(NOT TARGET tinyxml2)`. 

Both approaches seem to work locally for me.

clalancette:
I'm also running into issues building pluginlib and its dependencies on Ubuntu 18.04 because cmake 3.10 doesn't allow you to IMPORT libraries from other packages.  The solution in https://github.com/ros2/tinyxml2_vendor/pull/8 seems to fix it for me, along with removing the export from this package.  I'll go ahead and review https://github.com/ros2/tinyxml2_vendor/pull/8.

I also verified that with that fix in place and https://github.com/ros2/rosbag2/pull/403 in place, I was able to build rosbag2 with no absolute paths in it (except for yaml_cpp_vendor, a different issue), and I was also able to build nav2_rviz_plugins (so the problem from #192 is still solved).

clalancette:
I tried this PR out along with https://github.com/ros2/tinyxml2_vendor/pull/8 and https://github.com/ros2/rosbag2/pull/403 .  That replicates the test I made when looking at #193 .

Unfortunately, with *this* PR, I fail to build.  Where I fail is in trying to build rviz_common, with errors like:

```
[74.522s] librviz_common.so: undefined reference to `tinyxml2::XMLNode::NextSiblingElement(char const*) const'
[74.523s] librviz_common.so: undefined reference to `tinyxml2::XMLDocument::LoadFile(char const*)'
[74.523s] librviz_common.so: undefined reference to `tinyxml2::XMLElement::Attribute(char const*, char const*) const'
[74.524s] librviz_common.so: undefined reference to `tinyxml2::XMLElement::GetText() const'
[74.524s] librviz_common.so: undefined reference to `tinyxml2::XMLDocument::XMLDocument(bool, tinyxml2::Whitespace)'
[74.524s] librviz_common.so: undefined reference to `tinyxml2::XMLNode::Value() const'
[74.525s] librviz_common.so: undefined reference to `tinyxml2::XMLNode::FirstChildElement(char const*) const'
[74.525s] librviz_common.so: undefined reference to `tinyxml2::XMLDocument::~XMLDocument()'
[74.525s] collect2: error: ld returned 1 exit status
[74.526s] make[2]: *** [rviz_common_display_test] Error 1
```

rviz_common doesn't itself use TinyXML2, so it must be picking up the dependency from pluginlib.  However, without pluginlib exporting the TinyXML2 dependency, there's nothing to link against.

That seems like some sort of bug, but I don't know what it is at the moment.  I'm going to suggest that we go ahead with #190 and https://github.com/ros2/tinyxml2_vendor/pull/8, which improves the situation.  Then we can continue debugging what is going on with downstream consumers.  @Karsten1987 what do you think?

Karsten1987:
But pluginlib does export the dependency? Looking at the CI job [here](https://ci.ros2.org/job/ci_linux/10642/consoleFull) rviz built correctly. 

Did you use https://github.com/ros2/rviz/pull/542/files?

clalancette:
> Did you use https://github.com/ros2/rviz/pull/542/files?

I had missed this one.  With that as well, things built here.

pluginlib/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ add_library(${PROJECT_NAME} INTERFACE)
 target_include_directories(${PROJECT_NAME} INTERFACE
   "$<INSTALL_INTERFACE:include>")
 ament_target_dependencies(${PROJECT_NAME} INTERFACE
-  ament_index_cpp class_loader rcutils rcpputils tinyxml2_vendor TinyXML2)
+  ament_index_cpp class_loader rcutils rcpputils TinyXML2)
```
dirk-thomas, line 25: The TinyXML2 interface target is only defined by the vendor package (see ros2/tinyxml2_vendor#8). So I fail to see how this can work in all the cases when the vendor package is not find_packaged before.

Karsten1987: `ament_target_dependencies` has nothing to do with `find_package`, does it? It ultimately calls `target_link_libraries`, isn't that right?

If so, this change makes sense as there is no target `tinyxml2_vendor::tinyxml2_vendor` or the equivalent `tinyxml2_vendor_LIBRARIES` which we could link against.
And I think there should not be one. The target to link against is `TinyXML2` and the vendor package only provides the `TinyXML2` and its respective `FindTinyXML2.cmake` file, but should not itself define a target.
From what I understood, the line below, `ament_export_dependencies(.. tinyxml2_vendor)` makes sure that a call to `find_package(tinyxml2_vendor)` will always happen. 

But I might oversee something here.

dirk-thomas: You are right. Let me move my comment to other changes line.

dirk-thomas: Never mind. This change in combination with the next line exporting both `tinyxml2_vendor` and `TinyXML2` makes the manual logic in the extra file obsolete.

dirk-thomas (approved):

clalancette (approved):

mjcarroll (approved):
I also had to use this to resolve similar issues with `image_transport`

PR #192, opened 2020-05-08
[ros2] Export tinyxml2 directly from pluginlib-extras.cmake

seanyen:
When I attempted to build `nav2_rviz_plugins`, I run into this CMake error:

```
CMake Error: install(EXPORT "nav2_rviz_plugins" ...) includes target "nav2_rviz_plugins" which requires target "tinyxml2_vendor" that is not in any export set.
CMake Generate step failed.  Build files cannot be regenerated correctly
```

After some digging, it is coming from this `pluginlib-extras.cmake`.

1. I either need to add `IMPORTED` to make `tinyxml2_vendor` as an interface imported library,
2. Or I need to directly define the `pluginlib_TARGETS` to be `tinyxml2` which is the target exported from `TinyXML2` project.

### DIFF
--- a/pluginlib/pluginlib-extras.cmake
+++ b/pluginlib/pluginlib-extras.cmake
@@ -45,4 +45,9 @@ find_package(tinyxml2_vendor REQUIRED)
 find_package(TinyXML2 REQUIRED)
 list(APPEND pluginlib_LIBRARIES ${TinyXML2_LIBRARIES})
 
-list(APPEND pluginlib_TARGETS tinyxml2)
+add_library(tinyxml2_vendor INTERFACE IMPORTED)
+target_include_directories(tinyxml2_vendor INTERFACE
+  ${TinyXML2_INCLUDE_DIRS})
+target_link_libraries(tinyxml2_vendor INTERFACE
+  ${TinyXML2_LIBRARIES})
+list(APPEND pluginlib_TARGETS tinyxml2_vendor)
--- a/pluginlib/pluginlib-extras.cmake
+++ b/pluginlib/pluginlib-extras.cmake
@@ -45,9 +45,4 @@ find_package(tinyxml2_vendor REQUIRED)
 find_package(TinyXML2 REQUIRED)
 list(APPEND pluginlib_LIBRARIES ${TinyXML2_LIBRARIES})
 
-add_library(tinyxml2_vendor INTERFACE)
-target_include_directories(tinyxml2_vendor INTERFACE
-  ${TinyXML2_INCLUDE_DIRS})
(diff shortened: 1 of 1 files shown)


seanyen:
@dirk-thomas This is ready for review and merge. Thanks.

pluginlib/pluginlib-extras.cmake
```diff
@@ -45,9 +45,4 @@ find_package(tinyxml2_vendor REQUIRED)
 find_package(TinyXML2 REQUIRED)
 list(APPEND pluginlib_LIBRARIES ${TinyXML2_LIBRARIES})
 
-add_library(tinyxml2_vendor INTERFACE)
-target_include_directories(tinyxml2_vendor INTERFACE
-  ${TinyXML2_INCLUDE_DIRS})
-target_link_libraries(tinyxml2_vendor INTERFACE
-  ${TinyXML2_LIBRARIES})
-list(APPEND pluginlib_TARGETS tinyxml2_vendor)
+list(APPEND pluginlib_TARGETS tinyxml2)
```
dirk-thomas: Looking at the CMake module https://github.com/ros2/tinyxml2_vendor/blob/c89337439ffe9b5123ca310af0d746eb9cbfa110/cmake/Modules/FindTinyXML2.cmake I don't you can safely assume that the target `tinyxml2` exists in all the cases. Therefore I think your use case needs a different fix.

dirk-thomas (commented):
@seanyen Please don't mention me on all PRs you create. Many repositories have dedicated maintainers which will review incoming tickets.

seanyen:
@dirk-thomas Sorry for this one, I was mentioning you because it seems to be broken after this PR #189, so figured you might have better understanding of the issue I run into.

pluginlib/pluginlib-extras.cmake
seanyen: Ah, ok I missed that part. I reworked my fix to simply decorate `tinyxml2_vendor` as `IMPORTED`, which also fixed the CMake error.

dirk-thomas:
CI builds with only FastRTPS and no tests:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10679/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6097/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8684/)
  * Unrelated CMake warnings
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10577/)

dirk-thomas:
Thanks for the patch.

PR #193, opened 2020-05-11
Don't export tinyxml2 internal details.

clalancette:
That should be done by tinyxml2 itself.  Also, this doesn't work
on earlier versions of cmake (3.10, in particular).

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

Along with https://github.com/ros2/tinyxml2_vendor/pull/8, this is an alternative to #190.  This seems to fix the problems when building with rosbag2 (along with https://github.com/ros2/rosbag2/pull/403), and also seems to work for building the nav2_rviz_plugins that originally motivated #192.

@Karsten1987 FYI.

### DIFF
--- a/pluginlib/pluginlib-extras.cmake
+++ b/pluginlib/pluginlib-extras.cmake
@@ -44,10 +44,3 @@ endif()
 find_package(tinyxml2_vendor REQUIRED)
 find_package(TinyXML2 REQUIRED)
 list(APPEND pluginlib_LIBRARIES ${TinyXML2_LIBRARIES})
-
-add_library(tinyxml2_vendor INTERFACE IMPORTED)
-target_include_directories(tinyxml2_vendor INTERFACE
-  ${TinyXML2_INCLUDE_DIRS})
-target_link_libraries(tinyxml2_vendor INTERFACE
-  ${TinyXML2_LIBRARIES})
-list(APPEND pluginlib_TARGETS tinyxml2_vendor)


Karsten1987:
for my understanding, why does #190 not work for the nav2_rviz_plugins?

dirk-thomas (approved):
LGTM assuming it passes on all platforms.

clalancette:
> for my understanding, why does #190 not work for the nav2_rviz_plugins?

I think it would, but it just seems that exporting this at all from pluginlib is the wrong way to go.  That is, if we merge this and ros2/tinyxml2_vendor#8, we get the same benefit but it is more "correct".

siddhya:
I tried this patch but I am getting the following error with this change. Any way to fix it? New to ros so maybe a rookie mistake.

```
--- stderr: rosbag2_storage
librosbag2_storage.so: undefined reference to `tinyxml2::XMLNode::Value() const'
librosbag2_storage.so: undefined reference to `tinyxml2::XMLNode::FirstChildElement(char const*) const'
librosbag2_storage.so: undefined reference to `tinyxml2::XMLDocument::XMLDocument(bool, tinyxml2::Whitespace)'
librosbag2_storage.so: undefined reference to `tinyxml2::XMLNode::NextSiblingElement(char const*) const'
librosbag2_storage.so: undefined reference to `tinyxml2::XMLDocument::LoadFile(char const*)'
librosbag2_storage.so: undefined reference to `tinyxml2::XMLDocument::~XMLDocument()'
librosbag2_storage.so: undefined reference to `tinyxml2::XMLElement::Attribute(char const*, char const*) const'
librosbag2_storage.so: undefined reference to `tinyxml2::XMLElement::GetText() const'
```

clalancette:
> I tried this patch but I am getting the following error with this change. Any way to fix it? New to ros so maybe a rookie mistake.

You also need ros2/tinyxml2_vendor#8

clalancette:
So I'm actually going to close this one out, since it is a weaker version of #190 (that was my mistake, I misunderstood what they both did).  We'll continue in #190 instead.

Issue #168, opened 2019-10-31
Errors building with XCode 11 on Mojave


nuclearsandwich:
Installing XCode from the app store currently brings XCode 11. However there is a long list of errors when building pluginlib with it on macOS Mojave. It's unclear from the error text whether upgrading to macOS 10.15 before using XCode 11 will be sufficient to resolve the issue and I don't have any Catalina machines to try on.

stderr log from a test build: https://gist.github.com/nuclearsandwich/307e2f33edf5ba4db2188443bcf39987

clalancette:
I do believe that this is fixed by #212, so I'm going to close this out.  Feel free to reopen if I turn out to be wrong.

PR #209, opened 2020-09-25
do not use experimental on clang>=10

Karsten1987:
I could not get pluginlib to compile with a homebrew llvm clang. Note this is different of AppleClang which is shipped by default on OSX. The filesystem libraries are marked explicitly unavailable until OSX 10.15.
That's the same problem as described here: https://answers.ros.org/question/338832/ros2-source-install-macos-10151-qt_gui_cpp-fails/

Don't have much more info about this though. Some information might be here:
https://stackoverflow.com/questions/42633477/macos-clang-c17-filesystem-header-not-found

Might be related to https://github.com/ros/pluginlib/issues/168 

As a general question, I am wondering though why we ship pluginlib with this custom logic. It anyways depends on `rcpputils` already, why not using their filesystem implementation?


clalancette:
Obsoleted by #212, so closing.

PR #214, opened 2021-02-08
change default behaviour to use c++fs if stdlib is not explicitly spe…

asobhy-qnx:
…cified for QNX

The default C++ library when using qnx compiler is libc++, so if stdlib compiler option is not explicitly specified we need to use FILESYSTEM lib to be c++fs. The change in this commit applies this.


asobhy-qnx:
@clalancette can we merge this?

clalancette:
> @clalancette can we merge this?

Sure, that works for me.

clalancette (approved):

PR #221, opened 2021-10-25
Move LibraryLoadExceptions down a level for more accurate error msgs

DLu:
The error message saying "maybe it has the wrong name or maybe the library doesn't exist" has always bugged me.

PR #228, opened 2021-12-27
[noetic] Move LibraryLoadExceptions down a level #221

DLu:
#221 for noetic

### DIFF
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -417,9 +417,10 @@ std::string ClassLoader<T>::getClassLibraryPath(const std::string & lookup_name)
 /***************************************************************************/
 {
   if (classes_available_.find(lookup_name) == classes_available_.end()) {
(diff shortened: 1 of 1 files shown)


pluginlib/include/pluginlib/class_loader_imp.hpp
```diff
@@ -417,9 +417,10 @@ std::string ClassLoader<T>::getClassLibraryPath(const std::string & lookup_name)
 /***************************************************************************/
 {
   if (classes_available_.find(lookup_name) == classes_available_.end()) {
-    ROS_DEBUG_NAMED("pluginlib.ClassLoader", "Class %s has no mapping in classes_available_.",
-      lookup_name.c_str());
-    return "";
+    std::ostringstream error_msg;
+    error_msg << "Could not find library corresponding to plugin " << lookup_name <<
+      ". Make sure the plugin description XML file has the correct name of the library.";
+    throw pluginlib::LibraryLoadException(error_msg.str());
```
sloretz, line 423: Making this method throw where it didn't previously seems like a significant change in a released distro. I could be convinced with a deep enough pre-release test though.

The folks working on `ros-o` might be interested in this though https://github.com/ros-o/pluginlib

PR #229, opened 2021-12-27
[melodic] Move LibraryLoadExceptions down a level #221

DLu:
#221 for melodic


clalancette:
I'm really concerned that this change is going to cause breakage to the downstream consumers of pluginlib.  Especially given Melodic's age, I'm worried about a change in semantics like this.  Unless this is fixing a critical bug in Melodic, I'd rather not take this.

DLu:
That's fair.

PR #231, opened 2021-12-27
[foxy] Move LibraryLoadExceptions down a level #221

DLu:
#221 for foxy


jacobperron:
Following the thread on the main PR (https://github.com/ros/pluginlib/pull/221#issuecomment-1065157698), I don't think backporting this change is worth the potential fallout.

PR #232, opened 2022-01-13
Remove redundant throw of a std::runtime_error

allenh1:
A `std::runtime_error` is immediately thrown after being caught, so it should be equivalent to just not catching it at all.


clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20097/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14543/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20816/)

PR #234, opened 2022-03-01
Documented possible exception being thrown in ClassLoader destructor.

remod:
The class member `class_loader::MultiLibraryClassLoader lowlevel_class_loader_` can throw an exception when being destroyed (which actually happened to me). This can be hard to fix, especially if one defines the `ClassLoader` instance to be `static`. In this case the instance is destroyed when the executable exits where you can't catch exceptions.

Documenting this behavior should help others to be careful. Also it is consistent with the documentation of the other methods.

### DIFF
--- a/pluginlib/include/pluginlib/class_loader.hpp
+++ b/pluginlib/include/pluginlib/class_loader.hpp
@@ -77,7 +77,11 @@ public:
     std::string attrib_name = std::string("plugin"),
     std::vector<std::string> plugin_xml_paths = std::vector<std::string>());
 
-  ~ClassLoader();
+  /// The destructor attempts to unload all libraries which are still loaded.
+  /**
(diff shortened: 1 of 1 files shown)


remod:
It looks like the pipeline failed due to connection issues. Will retry later.

```
10:00:09 E: Unable to locate package ros-noetic-catkin
10:00:09 E: Unable to locate package ros-noetic-class-loader
10:00:09 E: Unable to locate package ros-noetic-cmake-modules
10:00:09 E: Unable to locate package ros-noetic-rosconsole
10:00:09 E: Unable to locate package ros-noetic-roslib
10:00:09 'apt install' failed and likely requires 'apt update' to run again
```

remod:
@nuclearsandwich kind ping

ahcorde (approved):

PR #235, opened 2022-03-01
Expose on-demand library load/unload in class loader constructor.

remod:
Currently, on-demand library loading/unloading is hard-coded to `false`. This leads to the fact that libraries are unloaded on destruction of the `ClassLoader` instance, which can throw exceptions (see https://github.com/ros/pluginlib/pull/234).

In my use case I'd like to unload the libraries on-demand, so I do not have to deal with exceptions when the destructor of the `ClassLoader` instance is called.

This PR exposes the setting in the `ClassLoader` constructor. The default value is `false` for backwards compatibility.


remod:
@nuclearsandwich kind ping

sloretz:
Thank you for the PR!

Unfortunately I don't think we should merge this one. It changes ABI, and ABI breaks cannot be accepted into released ROS distros.

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160).

PR #244, opened 2022-12-15
Avoid throwing on symlink loops.

iwanders:
Hi, we were seeing a throw from rviz with the following message:
```
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::status: Too many levels of symbolic links: "/nix/store/44kcmy9bsw44y13bf7x5kzlm9s4cdhpr-libgccjit-10.3.0/lib/lib/libcpr_viz.so"
Aborted
```
Here, the `/nix/store/44kcmy9bsw44y13bf7x5kzlm9s4cdhpr-libgccjit-10.3.0/lib/` directory is a symlink to itself, so there's an endless recursion. The `boost::filesystem::exists` method throws in this instance, but the exception is not caught anywhere, so rviz exits.

Instead, it is safer to use the overload that takes the error code and ensure that the function returned success, I don't think pluginlib should ever result in an uncaught throw?

fyi @mikepurvis, @jasonimercer, @nlunscher-cpr (CORE-23076)

### DIFF
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -436,10 +436,13 @@ std::string ClassLoader<T>::getClassLibraryPath(const std::string & lookup_name)
     it++)
   {
     ROS_DEBUG_NAMED("pluginlib.ClassLoader", "Checking path %s ", it->c_str());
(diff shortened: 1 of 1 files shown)


jasonimercer:
LGTM

ahcorde (approved):

PR #250, opened 2023-10-26
Fix wShadow compile warning

SteveMacenski:
`it` is defined on L479 and then again in the for loop. Renamed the for look items so that it doesn't compile with warnings turned to errors with `-wshadow` enabled

### DIFF
--- a/pluginlib/include/pluginlib/class_loader_imp.hpp
+++ b/pluginlib/include/pluginlib/class_loader_imp.hpp
@@ -488,12 +488,12 @@ std::string ClassLoader<T>::getClassLibraryPath(const std::string & lookup_name)
   RCUTILS_LOG_DEBUG_NAMED("pluginlib.ClassLoader",
     "Iterating through all possible paths where %s could be located...",
     library_name.c_str());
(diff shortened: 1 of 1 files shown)


SteveMacenski:
I see there is both a `rolling` and a `ros2` branch, so I selected Rolling but I wasn't sure which

clalancette:
> I see there is both a `rolling` and a `ros2` branch, so I selected Rolling but I wasn't sure which

Yeah, `rolling` is correct.  One of these days we'll make that clear.

clalancette (approved):
Yeah, good call.  I'll run CI on it next.

clalancette:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19689/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14195/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20419/)

SteveMacenski:
Sweet! The ticket I linked above can also be closed when this is merged. Found out after I ran into it Ryan reported it as well

PR #257, opened 2024-07-24
Fix runtime dependencies

at-wat:
Dependencies like class_loader are needed runtime. If packages using pluginlib are executed without class_loader installed, they fail with:
"Error loading shared library libclass_loader.so: No such file or directory"

Hit this problem on Alpine ROS which handles dependencies more strictly based on the package manifests.


sloretz (approved):
Thank you for the PR!

Issue #259, opened 2024-10-16
Feature Request: Python Plugins


DLu:
It's bugged me for years that there's no official Python bindings for this package. I'd like to be able to do the following in Python. 
 * Declare an "interface" (abstract class) in a ROS package.
 * Write an implementation of that interface in another ROS package with whatever metadata to link it to the first ROS package. 
 * List all implementations of the interface across the current ROS build environment.
 * Instantiate the implementation from a string, a la `loader.createSharedInstance("polygon_plugins::Square");`

The only "mainstream" version of this that I've seen is [rqt_bag](https://github.com/ros-visualization/rqt_bag/blob/master/rqt_bag/plugin.xml) which in turn uses [`rqt_gui.RospkgPluginProvider`](https://github.com/ros-visualization/rqt/blob/ros2/rqt_gui/src/rqt_gui/rospkg_plugin_provider.py). I've also written [a different implementation](https://github.com/MetroRobots/the_navigation_gauntlet/blob/6fd2f85904cee07076956f48d2df55bc5dfbaae0/navigation_metrics/navigation_metrics/metric.py) for my own use. 

However, I think there should be one common package for this, separate from `rqt_gui`. 

Would the maintainers be open to contributions along this line?

DLu:
![Screenshot 2024-10-16 at 15-33-47 Planning to Plan - New Flexible Navigation Interfaces - Google Slides](https://github.com/user-attachments/assets/5a52d345-c56b-449b-bfc7-cc3f30c76238)

Relevant slide from my 2018 ROSCon talk. When all you have is a hammer....

christophebedard:
Wouldn't the equivalent in Python be the built-in plugin system with `importlib`?

DLu:
I haven't done a ton with `importlib` itself. It would definitely be used for bullet 4 above (instantiation) but is there an `importlib` equivalent to listing all plugins?

christophebedard:
I'm pretty sure, yes.

I know that `colcon` uses it for its various plugin types. For example with its `package_selection` plugin:

1. For points 1, 3 and 4: this part of `colcon_core` retrieves all `package_selection` plugins using some utility wrapper around `importlib`: https://github.com/colcon/colcon-core/blob/8a08c158bc537b734a945b3ed84f756e9627d35b/colcon_core/package_selection/__init__.py#L98
[...]

PR #261, opened 2024-10-21
Python Based Pluginlib

DLu:
Implements #259. 

I learned a lot about entry points today. It ended up being way easier than I expected. 

After I made the initial issue, I did remember that `ros2cli` did something similar and that plus @christophebedard's [comment](https://github.com/ros/pluginlib/issues/259#issuecomment-2424738171) helped me figure all this out.

### DIFF
--- a/pluginlib/CMakeLists.txt
+++ b/pluginlib/CMakeLists.txt
@@ -45,6 +45,8 @@ install(
 
 install(DIRECTORY include/ DESTINATION include/${PROJECT_NAME})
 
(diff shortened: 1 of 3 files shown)


christophebedard:
Oh yeah, I don't know why I didn't think of `ros2cli`.

Since it's _that_ easy, does it belong here in `pluginlib`, though? Just asking.

DLu:
I concur that this is not needed here. I think it will be useful to still have the PR so that the next person who needs to know how to do this (which will likely be me) can just refer to the code here.

PR #262, opened 2024-10-22
License

### LINKED ISSUE #252: Add a License file
Please add a License file. The BSD license is stated in the package.xml, but for legal clarity, the license text should be included.

DLu:
Fixes #252 for ROS 2

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2009, Willow Garage, Inc.
+All rights reserved.
+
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Thanks for adding this.

I ended up changing the format somewhere here so that `ament_copyright` will be happy with it.  To be clear, we don't currently *use* `ament_copyright` in this package, but I checked that it would be happy with this change (I'll do a follow-up PR enabling ament_copyright).

With that, I'm going to approve and merge this (it can't possibly affect our CI builds, since it is currently outside of a package).

PR #284, opened 2025-10-23
Add dependency on ros2plugin in package.xml

mini-1235:
Hi, I think we need to add the dependency here so it is automatically installed when installing pluginlib

### DIFF
--- a/pluginlib/package.xml
+++ b/pluginlib/package.xml
@@ -29,6 +29,8 @@
   <depend>rcpputils</depend>
   <depend>tinyxml2</depend>
 
+  <exec_depend>ros2plugin</exec_depend>
+
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>


mini-1235:
@ahcorde Can you take a look here? Thanks

mini-1235:
friendly ping @ahcorde :)

What's your opinion here? I think this should be part of `ros_core`, we have `ros2cli_common_extensions` here 
https://github.com/ros2/variants/blob/c72de4fb5b6ecb908338e4780c570c8d7a5dc30c/ros_core/package.xml#L68
and `ros2bag` here  
https://github.com/ros2/rosbag2/blob/73b50297184d099f60d9d377844090bb87f405f8/rosbag2/package.xml#L15

mini-1235:
Oops, I didn't mean to close

ahcorde:
I'm not pretty sure about this one, we don't have a metapackage for `pluginlib` and adding this dependency will bring many python dependencies: `ros2cli`, `ros2pkg`, `rclpy`, etc.

@mjcarroll any thoughts on this one ?

mjcarroll:
I agree, this relationship seems inverted.  There are cases where you would want to use `pluginlib` without bringing in the dependencies required for command line tooling.

mini-1235:
@ahcorde @mjcarroll would it make sense to include it in the [variants](https://github.com/ros2/variants/blob/c72de4fb5b6ecb908338e4780c570c8d7a5dc30c/ros_core/package.xml#L76)￼ repository instead?

mjcarroll:
Either variants or https://github.com/ros2/ros2cli_common_extensions/blob/rolling/ros2cli_common_extensions/package.xml  I'm going to ask the PMC for opinions on this one.

mini-1235:
Thanks!

mini-1235:
> Either variants or https://github.com/ros2/ros2cli_common_extensions/blob/rolling/ros2cli_common_extensions/package.xml I'm going to ask the PMC for opinions on this one.

@mjcarroll Any updates :smile:

mini-1235:
I will close this one and continue the conversation to https://github.com/ros2/ros2cli_common_extensions

PR #291, opened 2026-02-20
Add support for passing arguments to constructors

pum1k:
# Description

This PR allows passing arguments to plugin constructors. This is done by exposing a new feature of class_loader added in ros/class_loader#223.

### DIFF
--- a/pluginlib/include/pluginlib/class_loader.hpp
+++ b/pluginlib/include/pluginlib/class_loader.hpp
@@ -34,6 +34,7 @@
 #include <string>
 #include <vector>
 
+#include "class_loader/interface_traits.hpp"
 #include "class_loader/multi_library_class_loader.hpp"
 #include "pluginlib/class_desc.hpp"
 #include "pluginlib/class_loader_base.hpp"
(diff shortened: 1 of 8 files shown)


ahcorde:
Pulls: ros/pluginlib#291
Gist: https://gist.githubusercontent.com/ahcorde/a60864c74e3698258ac6711f38798467/raw/f2aa7f4ba8de68858cc4a421df49836dcf859dbb/ros2.repos
BUILD args:  --packages-above-and-dependencies pluginlib
TEST args:  --packages-above pluginlib
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18275
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27764/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20794/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7625/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26757/)

SteveMacenski:
@ahcorde are you aware this breaks all existing uses of pluginlib? Currently Nav2 does not build due to this, which I highly suspect means that every plugin user would need to update their use of pluginlib to work (since we don't do anything fancy).

I think there either needs to be a workaround for objects without constructor arguments or a detailed migration guide for what updates need to be specifically made. Personally I think a library like this needs a tick-tock migration pattern to give folks time and awareness, but _feels_ to me like this could be done non-disruptively (though this is not my area of expertise)

pum1k:
@SteveMacenski the Nav2 test is broken, because it depends on **private** parts of the pluginlib. Code calling the public APIs of pluginlib requires no change.

SteveMacenski:
Huh - OK - false alarm. Still need to figure out a solution, but feel free to ignore me then :upside_down_face: 

I thought I saw this show before, but turns out I'm watching a different channel

PR #305, opened 2026-08-05
Fix pluginlib_enable_plugin_testing() docstring pitfalls.

clalancette:
The usage example wrapped the APPEND_ENV value in double quotes.  CMake passes those quotes through as literal characters in the environment variable value, so tests copying the example end up with a mock install path that is never found at test time.  Fix the example and call the pitfall out explicitly.

Also document that keeping the default PACKAGE_NAME shadows the real package when it is installed in a prefix which comes earlier in AMENT_PREFIX_PATH (as is the case when a package's own tests run against its install space), in which case pluginlib fails to find the plugin library in the mock install.

Also see https://github.com/ros2/urdf/pull/57 , which is where this was found.

I used Claude Fable 5 while creating this change.

### DIFF
--- a/pluginlib/cmake/pluginlib_enable_plugin_testing.cmake
+++ b/pluginlib/cmake/pluginlib_enable_plugin_testing.cmake
@@ -39,6 +39,9 @@ set(PLUGINLIB_ENABLE_PLUGIN_TESTING_DIR "${CMAKE_CURRENT_LIST_DIR}")
 #
 # CMake macros provided by ament_cmake for creating tests have an argument
 # called APPEND_ENV that should be used for modifying `AMENT_PREFIX_PATH`.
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30046/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22813/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/9931/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/28965/)

ahcorde:
https://github.com/Mergifyio backport lyrical kilted jazzy humble

PR #306, opened 2026-08-21
Fix pluginlib_enable_plugin_testing() docstring pitfalls. (backport #305) 

ahcorde:
The usage example wrapped the APPEND_ENV value in double quotes.  CMake passes those quotes through as literal characters in the environment variable value, so tests copying the example end up with a mock install path that is never found at test time.  Fix the example and call the pitfall out explicitly.

Also document that keeping the default PACKAGE_NAME shadows the real package when it is installed in a prefix which comes earlier in AMENT_PREFIX_PATH (as is the case when a package's own tests run against its install space), in which case pluginlib fails to find the plugin library in the mock install.

### DIFF
--- a/pluginlib/cmake/pluginlib_enable_plugin_testing.cmake
+++ b/pluginlib/cmake/pluginlib_enable_plugin_testing.cmake
@@ -39,6 +39,9 @@ set(PLUGINLIB_ENABLE_PLUGIN_TESTING_DIR "${CMAKE_CURRENT_LIST_DIR}")
 #
 # CMake macros provided by ament_cmake for creating tests have an argument
 # called APPEND_ENV that should be used for modifying `AMENT_PREFIX_PATH`.
+# The APPEND_ENV value must not be wrapped in double quotes; they would be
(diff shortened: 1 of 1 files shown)


ahcorde:
Pulls: ros/pluginlib#306
Gist: https://gist.githubusercontent.com/ahcorde/056193b8081aafd266b2d9339cfd92f8/raw/d57cba1580d81db0186b11f5003f3d6d49aaeff4/ros2.repos
BUILD args:  --packages-above-and-dependencies pluginlib
TEST args:  --packages-above pluginlib
ROS Distro: lyrical
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/20151
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/30219/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/22916/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/10030/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/29078/)

